PR #3, opened 2020-01-04
Proof of concept for generating dockerfiles automatically

noamross:
This is a proof of concept that generates versioned Dockerfiles based on data pulled from the [versions grid](https://docs.google.com/spreadsheets/d/11EHcl5i017ZWn__-PAPfT1d2GiNTGQbcfMuqyfkt-Rg/edit#gid=0).  `make-dockerfiles.sh` pulls that as CSV and produces the Dockerfiles using `Dockerfile.template`. The grid now includes things like ports to expose and entrypoints.

This approach makes the images independent of each other, which to some degree simplifies the build approach. It also produces pretty understandable, self-contained Dockerfiles.  A disadvantage of the current approach is that images do not share layers and this increases build time and storage.  Because the environment variables are set in bulk up top, images diverge at a relatively low level. We could try to set up the grid so that images still have cascading dependencies. Copying the scripts directory over also means that all images change when a single install script will change.

@cboettig


cboettig:
This is definitely interesting, thanks for working up the example.  I would definitely prefer we still generated "stacking" images though.  

Also, templating with Makefile and sed is cool and all, but is that really the standard practice for automatically generated Dockerfiles.  I mean, it seems everyone does automated Dockerfiles from templates these days, usually in super-flexible ways that support stacked images. e.g. looks like the tensorflow team uses a `spec.yml` to specify the combinations, and a custom `assembler.py` script to generate the dockerfiles: <https://github.com/tensorflow/tensorflow/tree/master/tensorflow/tools/dockerfiles#contributing>  Not sure what nvidia/cuda is doing, but I think they are also be auto-generating the Dockerfiles for all the different flavors which still stack nicely on top of each other... I think long-term if we go the templating route, using a dedicated templating system would be easier to maintain, but maybe not?

PR #4, opened 2020-01-12
Whisker / stacked 

noamross:
This implementes a **whisker** / stacked approach in which Dockerfiles are build by assembling pieces in `partials/` and `{{variables}}` are replaced with those in the version grid.  Images now share layers by declaring the parent image in the [version grid](https://docs.google.com/spreadsheets/d/11EHcl5i017ZWn__-PAPfT1d2GiNTGQbcfMuqyfkt-Rg/edit#gid=0).

Using **whisker** certainly makes the templating logic much easier.  Now, however, we are depending on having R in the build environment.  While I was at it I started building the images with **stevedore**, but we can drop this.

My Make-fu is a little weak so I couldn't get the `Makefile` to do what I wanted if you want to take a look, @cboettig.


cboettig:
@noamross I'm still digesting the details here bu this looks really good to me!  I think we could merge this and then continue to build out the stack in this format and see how it goes.  

I'm fine with the stevedore bit, it can be handy for an R user with limited docker experience, but also it's nice that things work fine without it too.  

(I still have a bit to wrap my head around on the version matrix side, and at some point want to think about what the mechanism is for stuff liking building images that skip part of the stack  (e.g. r-ver + tidyverse) etc, but those are separate issues anyway. )

noamross:
It's easy to add an image that skips the stack just by adding a row to the grid and specifying the base image and the partials you want to add. But I don't think it's how *users* should do it.

cboettig:
@noamross Can I merge this?

partials/Dockerfile.r-ver.partial
```diff
@@ -0,0 +1,27 @@
+# Add the Rocker install scripts
+COPY scripts /rocker_scripts
```
cboettig: I still do think we should copy scripts over individually as needed, which would play nicely with caching.  

Copying them all here is kinda pointless since we still have to repeat this in the downstream Dockerfiles if we are making separate Dockerfiles anyway, right?

partials/Dockerfile.rstudio.partial
```diff
@@ -0,0 +1,10 @@
+# Install RStudio
+ENV S6_VERSION={{S6_VERSION}}
+ENV RSTUDIO_VERSION={{RSTUDIO_VERSION}}
+RUN /rocker_scripts/install_rstudio.sh
```
cboettig: um, does this work? there's no `COPY` in this Dockerfile...

partials/Dockerfile.r-ver.partial
noamross: We touched on this before.  Ot does play better with caching when developing, not as much once things are stable.  I think it's much better for users who are customizing to have all of these scripts available to them as long as they are building on top any `r-ver` or derived image.

partials/Dockerfile.rstudio.partial
noamross: See above.

PR #5, opened 2020-01-25
Feedback on deeping the stack

cboettig:
@noamross Mostly opening this PR for discussion purposes.  Builds from the `whisker` branch.  

- Proposes a `versions.json` format (literally just a `read_csv() %>% write_json()` translation of the current version table.  I don't expect you'll like this, but I think it's worth looking at the concrete example for discussion purposes.  (Note it also simplifies the make-dockerfiles.R logic slightly).  

(As I noted on slack, my main motivation  for this is that I don't see the table logic currently, since it gives the illusion that all versions are set separately for each image).

- I've had some trouble with the stevedore script pulling / using my existing (debian-based) `rocker/image:tag` images instead of building these things from the ground up.  I've tweaked the Makefile temporarily to work around it.  

- I've extended the stack through geospatial and `ml`.  I haven't added the version control through environmental variables -- in fact, the the wisker templates for everything here are still all static since I think we need to think through just what gets versioned in each of those images.  

- There's obviously more testing needed in the install* scripts to ensure they are order-independent and stand-alone. 

- Please take a look at the python configuration in particular.  There are env vars here that need to be set as ENV in the Dockerfiles so that these persist into the user's runtime environment.  I don't see how that process can be entirely moved into `.sh` scripts that the user can run later.

- yeah I've reverted to copying individual build scripts here so that I have caching, but it will be easy enough to drop that once we've got other stuff ironed out and we can prove that the portable `rocker_scripts` idea really works independent of wisker.


noamross:
1) Honestly, json vs. csv isn't that different in my mind.  Both formats aren't great for editing or understanding by reading.  I preferred CSV because I actually thought of the Google Sheet as the better-to-edit-and-intuit tabular format (which is why downloading from the sheet to CSV is part of the Makefile).  Possibly more important is making a reasonably interpretable _display_ table that helps people understand what is in each image.

2) Re: install scripts, yes.  What I am going to try today is some cross-sourcing with install guards for this purpose.

3) I think with whisker I can add a flag that will allow us to switch the two COPY approaches for dev and build workflows.

noamross:
Some things added here:

- Whether we copy the files in `r-ver` or individually in scripts is now controlled by the env var ROCKER_DEV_WORKFLOW ("1" to copy individual for faster developer iteration).
- All dockerfiles templates bracketed by "start" and "end" templates.  CMD/ENTRYPOINT/EXPOSE are all in the latter.

noamross:
Question with the `ml` example: Do we want ML built on `geospatial` by default? I know that's what you started with in the beta repository, but it is a beta repository.

noamross:
OK, I added shiny server in here (builds but haven't yet tested).  I'm not sure if I got the logging setup right, as the rocker shiny image and ADD_SHINY had some slight differences.  But I believe the current setup will work in tandem with RStudio or not.

cboettig:
okay, really cool.  Built the full stacks from scratch and started poking around to confirm some expectations, made sure I could connect he gpu instance to gpu.  More to poke around here but overall I'm really liking this. 

I think we may as well merge this before we get too deep on branch-of-a-branch-of-a-fork... thoughts?  (Then we can start filing issues against this stack for next tasks that need more flushing out, such as the control of versions in the python / ml stack which is currently hardwired somewhat crudely in the scripts.  We can also work out some to-do list of checks / version combinations we want to make sure build).  

I will try to do more testing and experimentation using this to build different stacks.  Yeah, I don't much care if `ml` is 'before' or 'after' geospatial by default. Currently I'm picturing configure a "stack" by writing a custom JSON file that describes the stack, analogous to the ones we already have.  I know we might not be on the same page on that and maybe it's not a good idea, but happy to discuss.

I'm planning to dogfood these images a bunch (currently been using my previous ubuntu/nvidia image on my gpu server anyway) to get a firmer impression of what I'm happy with and what I want to tweak.  

I may play with some GH actions + GH docker repo stuff to get some automated remote builds for the stacks, mostly because it will be a good test of the architectural design, even if we don't use that for deployment or CI.

PR #21, opened 2020-02-28
docker-compose based image building

cboettig:
This PR adds `write_compose.R` which provides the function `write_compose()` which takes a JSON file and writes the docker-compose file so we can build the whole stack with a `docker-compose build` and push with a `docker-compose push`.

No `purrr`, no `jsonlite`, just base R and `yaml` :-).   

I've updated the GH actions script for image building as well, that might need a little more work/testing.


cboettig:
🎉 looks like GH-Actions are okay with this too.  Look, we now have packages! https://github.com/noamross/rocker-versioned2/packages

Do need to extend this to a more / more comprehensive JSON files, this just tested with a core set.  But will go ahead and merge this because *progress*.

cboettig:
just a note, the `docker-compose push` step seems weirdly fragile (will fail on gh-actions, but triggering a rebuild it goes fine).  Not sure what is up, possibly api limits

PR #36, opened 2020-05-03
Shiny, tensorflow, latest-tags, collapsing env vars

noamross:
Still some things to tweak, but I have to do some other things this afternoon and wanted you to have the latest.  In this PR

- Shiny 3.6.3 and 4.0.0 stacks
- Tensorflow cpu and gpu stacks
- Now adding "latest: true" to any image in the stackfiles (local or global) will result in it getting a ":latest" in the compose files.
- ENV vars are now collapsed so they are set in a single image layer.

### DIFF
--- a/compose/binder.yml
+++ b/compose/binder.yml
@@ -10,3 +10,8 @@ services:
     build:
       context: ..
       dockerfile: dockerfiles/Dockerfile_binder_devel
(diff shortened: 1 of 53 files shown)


cboettig:
Still need to see why ml & shiny aren't building successfully on gh actions

PR #40, opened 2020-05-10
Initial work on building with Builkit and ccache

noamross:
This is a first pass on user Docker Buildkit and ccache to speed up builds.  It requires a recent docker install with experimental features turned on. Still checking that it all works

Running `make BUILDKIT_CACHE=1 target` does the following:
- Creates dockerfiles that have 
   -  headers to trigger extended features
   - `RUN` statements specifying cacheing
    - Added `ARG` statements specifying CCACHE location and to use a custom `Makevars.ccache` for package installation that uses ccache
  - Installs ccache in the `install_R*.sh` 

Some things to consider/improve
  - We could install ccache by default (makes it easier for people to use it later)
  - `apt` packages can be cached, see [here](https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/experimental.md#example-cache-apt-packages), I just want to understand how this interacts with our scripts and cleanup
 - The `Makevars` approach could be how to parallel compiles.  `Ncpu` installs multiple _packages_ in parallel.  One can compile C code in parallel within individual packages like this:

    ```
    NPROC=$(shell nproc)
    NPROC_MIN=$(shell $(($NPROC>4?4:$NPROC)))
    MAKEFLAGS+="-j $(NPROC_MIN)"
    ```  

    But we'd want to think about nesting parallelism of stacks/images/packages/compiled files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@ STACKFILES=$(wildcard stacks/*.json)
 STACKS=$(notdir $(basename $(STACKFILES)))
 COMPOSEFILES=$(addprefix compose/,$(addsuffix .yml,$(STACKS)))
 PUSHES=$(addsuffix .push,$(STACKS))
+BUILDKIT_CACHE=0
 
 .PHONY: clean build setup push
 .PHONY: $(STACKS) $(PUSHES)
@@ -16,7 +17,7 @@ $(COMPOSEFILES): make-dockerfiles.R write-compose.R $(STACKFILES)
 build: $(STACKS)
 
 $(STACKS): %: compose/%.yml
-	docker-compose -f compose/$@.yml build
+	COMPOSE_DOCKER_CLI_BUILD=$(BUILDKIT_CACHE) DOCKER_BUILDKIT=$(BUILDKIT_CACHE) docker-compose -f compose/$@.yml build
 
 binder: geospatial
 
--- a/dockerfiles/Dockerfile_binder_4.0.0
+++ b/dockerfiles/Dockerfile_binder_4.0.0
@@ -7,6 +7,7 @@ LABEL org.label-schema.license="GPL-2.0" \
 
 
 
+
 RUN /rocker_scripts/install_python.sh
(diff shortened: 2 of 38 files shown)


noamross:
I adding cacheing of `apt` packages.  I believe the way I configured it, it should not bloat the image, nor change the default behavior users experience.  (That is, apt caching options are in a config file only used when an `APT_CONFIG` is set, and that is set via ARG so shouldn't carry through.)

noamross:
Noting for later: saving cache across github actions https://stackoverflow.com/a/60485339

noamross:
Hmm, nothing broken here but I can't seem to get it to actually _use_ the cache.  @jameshowison, given your experiments at https://github.com/howisonlab/test_repo_buildx_renv and question at https://github.com/docker/buildx/issues/255,  I wonder if you have any thoughts on how to debug this? [Here's one example Dockerfile](https://github.com/rocker-org/rocker-versioned2/blob/buildkit-cache/dockerfiles/Dockerfile_r-ver_4.0.0). Key differences from your setup to this are:

- We're building using [docker-compose's Buildkit interface](https://www.docker.com/blog/faster-builds-in-compose-thanks-to-buildkit-support/)
- We're not using `renv` but `ccache` for compiled code as well as cacheing `apt` packages

But mostly it's the same thing.  If you found anything useful for inspecting the state of the cache, I'd love to know!

jameshowison:
I never did find anything useful to see the contents, which is a bit weird, one should be able to mount them somehow! The buildx people seemed reasonably responsive, but I don't think I expressed myself well enough. Did you follow over to here: https://github.com/docker/buildx/issues/244  The trick in getting a tar file out showing the cache appears to be to do with the `-o` flag to `buildx` I couldn't get that to complete, perhaps something about recursive simlinks tripping it up?

is the APT cache working? I don't know anything about ccache. How does ccache know to look at `/root/.ccache`?

Is it clear that the buildx is working (showing the new progress format etc)?

PR #43, opened 2020-05-13
use RSPM for binary builds when available

cboettig:


cboettig:
Note that:

- it looks like binaries aren't available for focal, though it's possible the older binaries might work (usually I believe older binaries are compatible with platforms running newer libs and compilers, but not vice-versa).  So I've gone and dropped in the `bionic` link here anyway.  Should obviously be updated once they have an endpoint for `focal`
- Similarly, 4.0.0 does not yet have binaries available.  It should just install from source in those cases, and start building binaries once available.

noamross:
In the idea of the principle of least surprise, I thought that RSPM should not be the default CRAN for users, even if we build it.  As such, my commits switch it so that we use the CRAN/MRAN from before, but set an `RSPM_URL` environment variable. When you `make USE_RSPM=1` (the default), an alternate `Rprofile.site` is used at build time that sets this as CRAN for installs as well as setting the Http user-agent.

slopp:
@cboettig  perhaps not the best place to drop this comment, let me know what you'd prefer, but 4.0.0 binaries are now available. 

We'll be working on adding support for focal in our next release!

cboettig:
@slopp here works for me.  (hehe apparently we wound up having our discussion threads in PRs instead of issues threads, oh well).  


 and yes, we've been using the 4.0.0 images!  

Re `focal`, actually,  I tested the `bionic` repo on `focal` and it seems to work just as well.  Witness: 

```
docker run --rm -ti rocker/tidyverse:4.0.0 bash
root@9d19d4d71a4e:/# cat /etc/os-release | grep UBUNTU_CODENAME
UBUNTU_CODENAME=focal

root@9d19d4d71a4e:/# echo $CRAN
https://packagemanager.rstudio.com/all/__linux__/bionic/latest

root@9d19d4d71a4e:/# install2.r stringi
trying URL 'https://packagemanager.rstudio.com/all/__linux__/bionic/latest/src/contrib/stringi_1.4.6.tar.gz'
Content type 'binary/octet-stream' length 2471760 bytes (2.4 MB)
==================================================
downloaded 2.4 MB

* installing *binary* package ‘stringi’ ...
* DONE (stringi)
```

binary `stringi` on `focal`, :tada:

As I mentioned in the email thread, I believe the practice of building images on older linux releases for deployment on newer ones is the standard strategy of python / manywheels, so I think this shouldn't be disastrous, though no doubt we'll all thank you for building newer versions (maybe I overestimate progress in compilers, but I'm amazed python community is happy with binaries built on compilers that are a decade old!)

The transitions has been very smooth! 

A few obvious / expected gotchas (e.g. the image with ubuntugis-unstable needs to install from source instead; likewise `littler` from binary is unhappy since our `Rlib.so` is is in `/usr/local/lib` since we compiled R from source...)

I think we are still waiting for a good resolution for the issue with RStudio's `remotes::install_dev) ` being discussed in #51, sounds like you're already on top of that and it's rather an edge case anyhow. 

Thanks again for the great work on this @slopp and team!

PR #45, opened 2020-05-13
Fixup install_shiny_server script and dependency scripts

manselmi:
Hello, and thank you for developing and maintaining the `rocker` R images.  :)

Past versions of the `rocker/rstudio` image had a `/etc/cont-init.d/add` script that made it easy to automatically install and configure Shiny Server. I recently noticed that this script has been replaced by `/rocker_scripts/install_shiny_server.sh`, so I tried running it.

However, the dependency script `install_pandoc.sh` failed due to assumptions that the script would only be one run once. I've made minimal changes to allow the script to be re-run and trimmed whitespace.

Also, I noticed a bug in `install_s6init.sh` where the string literal `S6_VERSION` was being written to `/rocker_scripts/.s6_version` instead of the value of the `S6_VERSION` variable. I've fixed that as well, and also trimmed whitespace.

I also addressed several issues with `install_shiny_server.sh`, including fixing `shiny.shiny` → `shiny:shiny` in a `chown` command, replacing a multi-line `echo` command with some unexpected substitutions with `cat`-ing a heredoc, and removing a bare stdout redirection operator that had no destination specified. Also, I trimmed whitespace. :)

After making these changes, I was able to get Shiny Server working within `rocker/rstudio`.

### DIFF
--- a/scripts/install_pandoc.sh
+++ b/scripts/install_pandoc.sh
@@ -8,7 +8,7 @@ PANDOC_VERSION=${1:-${PANDOC_VERSION:-default}}
 
 apt-get update && apt-get -y install wget
 
-if [-x "$(command -v pandoc)" ]; then
+if [ -x "$(command -v pandoc)" ]; then
   INSTALLED_PANDOC=$(pandoc --version 2>/dev/null | head -n 1 | grep -oP '[\d\.]+$')
 fi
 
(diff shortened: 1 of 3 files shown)


manselmi:
I don't have permission to tag anyone as a reviewer, so FYI: @cboettig @noamross

Sorry if you had already been notified about this PR; I'm not sure what GitHub's default notification settings are.

cboettig:
Thanks @manselmi this looks great.  Much appreciated.

Issue #50, opened 2020-05-19
Downstream effects from switching the CRAN env var to https://packagemanager.rstudio.com


lcolladotor:
Hi,

I've noticed that GitHub Actions (GHA) workflows that depend on the rocker docker images have had a bit of a hiccup lately with `remotes::install_dev("pkgdown")` as I reported at https://github.com/r-lib/remotes/issues/501. It seems like the root issue was the recent change at https://github.com/rocker-org/rocker-versioned2/commit/2dcd686abe67040e88ff1317cfc4bc59bb5a7338 which switched the `CRAN` environment variable from `https://cran.r-project.org` to `https://packagemanager.rstudio.com/all/__linux__/bionic/latest` in most cases. Jim Hester at https://github.com/r-lib/remotes/issues/501#issuecomment-630846911 suggests that `options(repos = c(RSPM = "xyz", CRAN = 'abc'))` is the better way to use the RSPM, so would it be ok to:

* revert the `CRAN` environment variable?
* set the `RSPM` environment variable to `https://packagemanager.rstudio.com/all/__linux__/bionic/latest/` (like you had done already)?

I ignore if there were other reasons why you changed `CRAN`, but if you want, I can send a PR that does these 2 changes.

Best,
Leo

lcolladotor:
Hi,

Given that resolving this issue (in appearance) seemed straight forward, I just sent a PR with the suggested code changes.

Best,
Leo

cboettig:
@lcolladotor given discussion in #51 I think we can close this?

lcolladotor:
Yes, based on the promise of adding support for `remotes::install_dev()` in the RSPM noted at https://github.com/rocker-org/rocker-versioned2/pull/51#issuecomment-634694250. Plus the `remotes::install_github()` temporary workaround.

Thanks again for looking into this.

PR #51, opened 2020-05-19
Resolve https://github.com/rocker-org/rocker-versioned2/issues/50

lcolladotor:
As discussed at https://github.com/rocker-org/rocker-versioned2/issues/50 and https://github.com/r-lib/remotes/issues/501, for using the RStudio Package Manager, it seems like the best option is to set the `RSPM` environment variable instead of changing the `CRAN` one. This pull request reverts https://github.com/rocker-org/rocker-versioned2/commit/2dcd686abe67040e88ff1317cfc4bc59bb5a7338 and adds the RSPM environment variable.


noamross:
Thanks, @lcolladotor.   I like this approach. @cboettig, do you agree?

This also needs the repo to be set in `scripts/install_R.sh` and `scripts/install_R-devel.sh`, like so:

```
## Add a default CRAN mirror
if [[ -n "$RSPM" ]]; then
  echo "options(repos = c(RSPM = '${RSPM}', CRAN = '${CRAN}'), download.file.method = 'libcurl')" >> ${R_HOME}/etc/Rprofile.site
else
  echo "options(repos = c(CRAN = '${CRAN}'), download.file.method = 'libcurl')" >> ${R_HOME}/etc/Rprofile.site
fi
```
Can you do so or make your branch editable by maintainers?

(@cboettig, now that we moved back to building R from source, do we need `install_R-devel.sh` anymore?)

lcolladotor:
Hi Noam,

I do have the "Allow edits by maintainers" checkbox ticked, or is it something else that I'm missing? I'm still new to sending pull requests 😅

Screen Shot 2020-05-19 at 12 42 19 PM

In any case, I just pushed a commit that implements the code changes you suggested at https://github.com/rocker-org/rocker-versioned2/pull/51#issuecomment-630939332.


Best,
Leo

cboettig:
@lcolladotor thanks for filing this, definitely still working out the kinks here.  Still, I'd love to get a better understanding of the details first. 

From the RSPM webpage and discussions with @slopp, I understood that we could use RSPM snapshots as an alternative to MRAN in order to get version-stable installs.  I don't understand how that squares with @jimhester 's [comment](https://github.com/r-lib/remotes/issues/501#issuecomment-630846911) 

> In general if you want to use the package manager it should set in conjunction with a regular CRAN repo, e.g. options(repos = c(RSPM = "xyz", CRAN = 'abc')) or similar.

It seems there's another underlying issue here in how RSPM mirror differs from a typical CRAN mirror in a way that, say, MRAN does not, which is causing this side-effect?

noamross:
> It seems there's another underlying issue here in how RSPM mirror differs from a typical CRAN mirror in a way that, say, MRAN does not, which is causing this side-effect?

Yes, that's right. RSPM doesn't have some files of the package available under the path to binaries that way CRAN or MRAN do on the path to sources.  For instance, https://packagemanager.rstudio.com/all/__linux__/bionic/latest/web/packages/pkgdown/ doesn't have the typical package landing page nor the DESCRIPTION at https://packagemanager.rstudio.com/all/__linux__/bionic/latest/web/packages/pkgdown/DESCRIPTION

lcolladotor:
No worries Carl =) I'm also trying to understand the details, so I'm learning from this experience and Noam's reply was great ^^. Thanks!

slopp:
@noamross I  am not sure I follow your comment. RSPM doesn't server the package web pages, but it does meet all the of the requirements we are aware of for install.packages and friends, and would be a superset of MRAN/CRAN from that perspective. Is there are reason those files are important for a programmatic client? 

I don't think you'd need to set a CRAN fall back - certainly not in the "snapshot" case where that could cause confusion (newer packages coming in from the fallback). Even with MRAN you would see discrepancies between snapshots. If you are in the snapshot case I'd recommend a single RSPM repo with a versioned repo ID. For the "latest" case, the only reason you'd want a CRAN fall back to RSPM is if you are sensitive to the 1-2 day delay that can exist between a package being available on CRAN and being available on RSPM.

noamross:
@slopp The **remotes** (and thus **devtools**) package queries DESCRIPTION files that are served on CRAN but not appear to be served by RSPM: https://github.com/r-lib/remotes/blob/9b5dc29102a486df2f42c88bb19027a7cd54a721/R/install-dev.R#L3-L6

slopp:
We will look into adding support for these paths and `remotes::install_dev`. If the main use case for this workflow is to install the latest pkgdown site, perhaps we could modify the call and use `remotes::install_github` directly as an alternative workaround - though I am not clear on all the places where `remotes::install_dev` is regularly referenced

lcolladotor:
Hi,

I just switched from `remotes::install_dev()` to `remotes::install_github()` yesterday https://github.com/lcolladotor/biocthis/commit/3b96da142aef74df69f35664a13d1283359d5db5 ^^.

As for where `remotes::install_dev()` is used, I have no idea either.

Best,
Leo

cboettig:
still waiting for RSPM support on focal, but will implement then.  It does sound like RSPM should be valid as `CRAN` repo for the reasons discussed.  `remotes::install_dev` seems to be operating on an undocumented assumption.

PR #64, opened 2020-06-24
Add README

noamross:
First pass at a top-level README


cboettig:
:+1:  This looks great.  

RSPM isn't available yet for focal.  

Can you add a bit more about which Ubuntu we use?  Specifically:
> Images are now based on Ubuntu LTS releases.  Images will use the most recent LTS available at the time when the corresponding R version was released.  Thus all 4.0.0 images are based on Ubuntu 20.04.

Issue #69, opened 2020-07-13
rstudio:4.0.2-ubuntu18.04 ?


nturaga:
Hi @cboettig 

Is the rstudio:4.0.2 - ubuntu18.04 image going to be made available? I can't seem to find it in the stack of Dockerfiles.

Best,

Nitesh

eddelbuettel:
Can you migrate the BioC stack to 20.04 which, as you know, is the current LTS?

nturaga:
Hi @eddelbuettel The bioconductor build system is currently on 18.04 and it would require significant effort to migrate our build system to 20.04 at the moment. This is the reason we chose to use 18.04 in the first place so that there is more symmetry between the bioconductor_docker image and the bioconductor build system.

PR #70, opened 2020-07-13
Add r-ver:4.0.2-ubuntu18.04 and rstudio:4.0.2-ubuntu:18.04

nturaga:
These images are used by the bioconductor images and it would be really
helpful to the whole Bioconductor community if they are included in the
rocker-versioned2 repository

This addresses #69

### DIFF
--- a/dockerfiles/Dockerfile_r-ver_4.0.2-ubuntu18.04
+++ b/dockerfiles/Dockerfile_r-ver_4.0.2-ubuntu18.04
@@ -1,0 +1,24 @@
+FROM ubuntu:18.04
+
+LABEL org.label-schema.license="GPL-2.0" \
+      org.label-schema.vcs-url="https://github.com/rocker-org/rocker-versioned" \
+      org.label-schema.vendor="Rocker Project" \
+      maintainer="Carl Boettiger <cboettig@ropensci.org>"
+
+ENV R_VERSION=4.0.2
+ENV TERM=xterm
+ENV LC_ALL=en_US.UTF-8
+ENV LANG=en_US.UTF-8
+ENV R_HOME=/usr/local/lib/R
+ENV CRAN=https://packagemanager.rstudio.com/all/__linux__/bionic/latest
+ENV TZ=UTC
+
+COPY scripts /rocker_scripts
+
+RUN /rocker_scripts/install_R.sh
+
+
+CMD ["R"]
+
+
+
--- a/dockerfiles/Dockerfile_rstudio_4.0.2-ubuntu18.04
+++ b/dockerfiles/Dockerfile_rstudio_4.0.2-ubuntu18.04
@@ -1,0 +1,21 @@
+FROM rocker/r-ver:4.0.2-ubuntu18.04
+
+LABEL org.label-schema.license="GPL-2.0" \
+      org.label-schema.vcs-url="https://github.com/rocker-org/rocker-versioned" \
+      org.label-schema.vendor="Rocker Project" \
+      maintainer="Carl Boettiger <cboettig@ropensci.org>"
+
+ENV S6_VERSION=v1.21.7.0
+ENV RSTUDIO_VERSION=latest
+ENV PATH=/usr/lib/rstudio-server/bin:$PATH
+
+
+RUN /rocker_scripts/install_rstudio.sh
+RUN /rocker_scripts/install_pandoc.sh
+
+EXPOSE 8787
+
+CMD ["/init"]
+
+
+
--- a/stacks/core-4.0.2-ubuntu18.04.json
+++ b/stacks/core-4.0.2-ubuntu18.04.json
@@ -1,0 +1,54 @@
+{
+  "ordered": true,
+  "TAG": "4.0.2-ubuntu18.04",
+  "LABEL": "org.label-schema.license=\"GPL-2.0\" \\\n      org.label-schema.vcs-url=\"https://github.com/rocker-org/rocker-versioned\" \\\n      org.label-schema.vendor=\"Rocker Project\" \\\n      maintainer=\"Carl Boettiger <cboettig@ropensci.org>\"",
+
+  "stack": [
+  {
+    "IMAGE": "r-ver",
+    "FROM": "ubuntu:18.04",
+    "ENV": {
+      "R_VERSION": "4.0.2",
+      "TERM": "xterm",
+      "LC_ALL": "en_US.UTF-8",
+      "LANG": "en_US.UTF-8",
+      "R_HOME": "/usr/local/lib/R",
+      "CRAN": "https://packagemanager.rstudio.com/all/__linux__/bionic/latest",
+      "TZ": "UTC"
+    },
+    "COPY": "scripts /rocker_scripts",
+    "RUN": "/rocker_scripts/install_R.sh",
+    "CMD": "[\"R\"]"
+  },
+  {
+    "IMAGE": "rstudio",
+    "FROM": "rocker/r-ver:4.0.2-ubuntu18.04",
+    "ENV": {
+      "S6_VERSION": "v1.21.7.0",
+      "RSTUDIO_VERSION": "latest",
+      "PATH": "/usr/lib/rstudio-server/bin:$PATH"
+    },
+    "RUN": ["/rocker_scripts/install_rstudio.sh", "/rocker_scripts/install_pandoc.sh"],
+    "CMD": "[\"/init\"]",
+    "EXPOSE": 8787
+  },
+  {
+    "IMAGE": "tidyverse",
+    "FROM": "rocker/rstudio:4.0.2-ubuntu18.04",
+    "RUN": "/rocker_scripts/install_tidyverse.sh"
+  },
+  {
+    "IMAGE": "verse",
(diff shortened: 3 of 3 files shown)


cboettig:
Thanks, but as Dirk points out in #69, is there a reason that you need things on the old Ubuntu-18.04 (bionic) LTS, rather than the current Ubuntu 20.04 (focal) currently underlying `rocker/rstudio:4.0.2`?  

Also just a technical note, these Dockerfiles are now generated automatically from the `stacks` descriptions, e.g. https://github.com/rocker-org/rocker-versioned2/blob/master/stacks/core-4.0.2.json makes the Dockerfiles for `rocker/rstudio:4.0.2` and friends.

nturaga:
Hi @cboettig The bioconductor build system is currently on 18.04 and it would require significant effort to migrate our build system to 20.04 at the moment. This is the reason we chose to use 18.04 in the first place so that there is more symmetry between the bioconductor_docker image and the bioconductor build system.

mtmorgan:
The 18.04 image uses the RStudio package manager, but the RStudio package manager builds against R-4.0.2, so  for instance one sees 
```
$ docker run -it --rm rocker/r-ver:4.0.0-ubuntu18.04 R --quiet
> install.packages("Rcpp")
Installing package into ‘/usr/local/lib/R/site-library’
(as ‘lib’ is unspecified)
trying URL 'https://packagemanager.rstudio.com/all/__linux__/bionic/latest/src/contrib/Rcpp_1.0.5.tar.gz'
Content type 'binary/octet-stream' length 3238492 bytes (3.1 MB)
==================================================
downloaded 3.1 MB

* installing *binary* package ‘Rcpp’ ...
* DONE (Rcpp)

The downloaded source packages are in
	‘/tmp/RtmpDTZCyK/downloaded_packages’
> library(Rcpp)
Error: package or namespace load failed for ‘Rcpp’ in dyn.load(file, DLLpath = DLLpath, ...):
 unable to load shared object '/usr/local/lib/R/site-library/Rcpp/libs/Rcpp.so':
  /usr/local/lib/R/site-library/Rcpp/libs/Rcpp.so: undefined symbol: EXTPTR_PTR
In addition: Warning message:
package ‘Rcpp’ was built under R version 4.0.2
```

eddelbuettel:
Hm, but as far as I understand that issue `EXTPTR_PTR` is "purely" an upstream R bug between versions 4.0.0, 4.0.1 (both bad AFAIK) and 4.0.2 (good) and not related at all to 18.04 and 20.04.

cboettig:
@nturaga @mtmorgan thanks both for the feedback and details. 

Apologies for some confusion, note that the `rocker/r-ver:4.0.0-ubuntu18.04` tag isn't being actively maintained, I just haven't figured out how to delete tags yet.  In general we've pinned old versions, so r-ver:4.0.0 (on `20.04`), also has RSPM, but is pinned to the old RSPM snapshot and not floating to latest CRAN.  

If there are any issues that are specific to ubuntu 18.04 vs 20.04 it would be great to know about them.  As Dirk points out in the example above, that really looks like an issue regarding using old versions that is resolved by using the latest versions.  We could resurrect the 18.04 tag set if really necessary, but so far everything we've seen on 20.04 is quite backwards compatible with stuff that previously ran on 18.04.   Ideally we'd like to stick with supporting only the concurrent LTS and not also maintain separate tags for older LTSs if possible...

mtmorgan:
I'm mostly following, but it seems the 18.04 image is not correctly pinned to the RStudio package manager? I had done

```
rocker-versioned2 master$ docker build -f dockerfiles/Dockerfile_r-ver_4.0.0-ubuntu18.04 -t rocker/r-ver:4.0.0-ubuntu18.04 .
```
and get
```
> options("repos")
$repos
                                                            CRAN
"https://packagemanager.rstudio.com/all/__linux__/bionic/latest"
```
where the 'latest' path is built against 4.0.2 (today, anyway ;)). It seems instead that what I'd like is something like

```
> options(repos = c(CRAN="https://packagemanager.rstudio.com/all/278"))
```

which builds using the May 3 CRAN snapshot (about the time of the 4.0.0 release?) and installs Rcpp from source. My understanding is that the '278' identifier is not actually stable...

I guess you're saying that the 18.04 image isn't maintained, so the bad pin isn't going to be fixed...

Maybe the 'latest' RSPM path on the 4.0.2 images will be problematic in the future?

eddelbuettel:
I think what Carl is trying to explain is that 
 - if you want `versioned2` then you will be with 20.04 as that is all that the (all-volunteer, also active elsewhere) team has built
 - if you want 18.04 with RSPM you can get find it [here as a Dockerfile](https://github.com/rocker-org/rocker/tree/master/r-rspm/bionic) and [here as a container](https://hub.docker.com/repository/docker/rocker/r-rspm)  (and I can never remember what others can see; make the mistake of linking to "my" pages there all the time...)

```sh
edd@rob:~$ docker run --rm -ti rocker/r-rspm:18.04 bash
root@7a454a038e2d:/# install.r RcppArmadillo     # so that we get two, also install.r just wraps install.packages() ...
also installing the dependency ‘Rcpp’

trying URL 'https://packagemanager.rstudio.com/all/__linux__/bionic/latest/src/contrib/Rcpp_1.0.5.tar.gz'
Content type 'binary/octet-stream' length 3238492 bytes (3.1 MB)
==================================================
downloaded 3.1 MB

trying URL 'https://packagemanager.rstudio.com/all/__linux__/bionic/latest/src/contrib/RcppArmadillo_0.9.900.1.0.tar.gz'
Content type 'binary/octet-stream' length 2031849 bytes (1.9 MB)
==================================================
downloaded 1.9 MB

begin installing package ‘Rcpp’
* installing *binary* package ‘Rcpp’ ...
* DONE (Rcpp)
begin installing package ‘RcppArmadillo’
* installing *binary* package ‘RcppArmadillo’ ...
* DONE (RcppArmadillo)

The downloaded source packages are in
        ‘/tmp/downloaded_packages’
root@7a454a038e2d:/# 
```

mtmorgan:
Thanks for your patience; and for the hard work and valuable service that you, Carl and others provide. Also for taking a 'bold' step and providing RSPM.

What got us into trouble was this line https://github.com/rocker-org/rocker-versioned2/blob/93ded8614fc6edba8e646116bffabe9c33607ec2/dockerfiles/Dockerfile_r-ver_4.0.0-ubuntu18.04#L13  which is not pinned to an RSPM version, and is I think a bug.

I can see several paths forward now, so probably the issue should be closed.

eddelbuettel:
Doesn't RSPM get that from the user-agent header?  I am adding (example from bionic) in `Rprofile.site`:

```r
local({
    r <- getOption("repos")
    #r["CRAN"] <- "https://cloud.r-project.org"
    r["CRAN"] <- "https://packagemanager.rstudio.com/all/__linux__/bionic/latest"
    options(repos = r)
    options(HTTPUserAgent = sprintf("R/%s R (%s)", getRversion(), paste(getRversion(), R.version$platform, R.version$arch, R.version$os)))
})
```

cboettig:
@mtmorgan yes, correct. A bit more precisely, it's this line, which creates said line in the Dockerfile:

https://github.com/rocker-org/rocker-versioned2/blob/93ded8614fc6edba8e646116bffabe9c33607ec2/stacks/core-4.0.0-ubuntu18.04.json#L16

which was never frozen after 4.0.1 was released, because there were no plans maintaining the bionic based images.  I probably should have deprecated the `core-4.0.0-ubuntu18.04` stack and Dockerfile at that time, but I didn't do so, my apologies.  In any event like you say the ubuntu18.04 tag has not been maintained and it seems that is causing issues.  

anyway I think your experience should be better if you try building on the standard `4.0.0` or `4.0.2` tags.  RStudio recently released support for RSPM on 20.04, so those ones will now have binary builds as well, and should in general behave much better.

p.s. yup, as Dirk shows, RSPM gets the correct version of R from the HTTPUserAgent header, while the ubuntu flavor and the choice to have either `latest` or a pinned snapshot are controlled by setting the CRAN URL.  (Also to request source instead of binary, you have to edit the CRAN url, just dropping the `__linux__` part.  AFAIK the RSPM snapshot numbers are indeed stable, if you find otherwise I think that would be a bug we should report to RStudio.

Will close this out but feel free to continue discussion.  Sounds like I should also go ahead and deprecate those Dockerfiles that we aren't maintaining!

mtmorgan:
Thanks for the very helpful discussion.

On the 4.0.0 image above (https://github.com/rocker-org/rocker-versioned2/pull/70#issuecomment-657735381) loading Rcpp results in
```
In addition: Warning message:
package ‘Rcpp’ was built under R version 4.0.2
```
Conversely, using the 4.0.2 image and a pinned repository such as
```
options(repos = c(CRAN = "https://packagemanager.rstudio.com/all/__linux__/bionic/291"))
```
gets the 4.0.0 binary (4.0.2 didn't exist at that date, so perhaps I wasn't expecting a binary installation...)
```
> packageDescription("Rcpp")$Built
[1] "R 4.0.0; x86_64-pc-linux-gnu; 2020-05-09 16:15:14 UTC; unix"
> getRversion()
[1] ‘4.0.2’
```
I infer that the minor but not patch version is used by the RStudio package manager to select binaries...

PR #87, opened 2020-09-24
Update README.md

jvanlunenburg:
RStudio and RStudio Server are two different products with a similar interface! Please stop conflating them in the Rocker Project.


eddelbuettel:
Thanks. And if you pay that much attention to minor details [1] you may as well get your capitalization right. So I permitted myself to correct your one-liner in three places.

[1] Which are largely irrelevant because how could you possibly ship RStudio Desktop in a container?

jvanlunenburg:
Hah thanks, you may think it pedantic but I found it confusing as I actually downloaded a desktop version container by accident.

PR #91, opened 2020-10-02
Initial Kubeflow support

DavidSpek:
If the environment variable `NB_PREFIX` is used, it is assumed rstudio is from kubeflow and thus userconf.sh will set `www-root-path` to
 `$NB_PREFIX`, set `www-port=8888` and it will disable auth. `RSTUDIO_VERSION=daily` or `RSTUDIO_VERSION=preview` need to be used (daily not working, preview not tested). On daily, the 404 page loads saying:
```
RSudio Server
The requested page was not found.
Please contact your system administrator for assistance, or click here to go back.
```
This was tested with a Dockerfile that exposes port 8888 rather than 8787 (Kubeflow expects 8888).

### DIFF
--- a/scripts/install_rstudio.sh
+++ b/scripts/install_rstudio.sh
@@ -88,6 +88,15 @@ echo "lock-type=advisory" > /etc/rstudio/file-locks
 cp /etc/rstudio/rserver.conf /etc/rstudio/disable_auth_rserver.conf
 echo "auth-none=1" >> /etc/rstudio/disable_auth_rserver.conf
 
+## Prepare optional configuration file to disable authentication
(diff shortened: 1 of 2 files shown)


cboettig:
@DavidSpek This looks reasonable to me, do you still want this labeled as draft or shall we move forward?

DavidSpek:
@cboettig Let me go over this again quickly. We now have an RStudio Dockerfile for Kubeflow which took a lot of work to get everything working correctly with conda and such. There might be some improvements we found while working on that which will be useful here as well.

cboettig:
awesome, thanks, much appreciated

PR #92, opened 2020-10-21
Install sqlite3

Robinlovelace:
For https://github.com/rocker-org/geospatial/issues/31


cboettig:
nice work! will give this a whirl...

Robinlovelace:
Thanks Carl. Apologies for the duplication in https://github.com/rocker-org/rocker-versioned2/pull/92/files#diff-b6f3273ba43db16540dc5558127e509b39c3d51e187a963519da5228920d9f10R167-R174 but I tried both and only the latter worked. Feel free to clean-up, look forward to seeing if this gets the `osgeo-dev` image going - does than need the rest of the 'geospatial stack' and spatial packages installed though?: https://github.com/rocker-org/rocker-versioned2/blob/master/dockerfiles/Dockerfile_geospatial_devel

PR #97, opened 2020-12-04
Update dev_osgeos.sh

Robinlovelace:


cboettig:
whoops, missed this! but done now.  Also mapped over to the stack file (and I renamed the script to be `dev_osgeo.sh` since the tag is `dev-osgeo` with no `s` on the end.  

Anyway, thanks for this!

PR #111, opened 2021-01-30
bugfix for `install_python.sh`

chrismazanec:
bugfix; sourced `install_python.sh` so vars are accessible from include


cboettig:
good catch, thanks!

PR #116, opened 2021-02-09
Tidy build scripts

damacus:
- Fixes missing shebangs in some places
- Removes empty lines and spaces

This is more of a preemptive PR for those editors out there that will prune white space where they see it. For example VSCode will remove spaces in files that you touch and I'm planning to make a couple of PRs in the future.


eddelbuettel:
Thank you for taking the time to prepare a PR.  Generally speaking it may be prudent to first discuss in any issue ticket if there was an appetite for _41 file_ PR.   

At a minimum it will essentially destroy commit histories and git blame.  For that and other reasons I tend to say "Thanks, but no thanks" in my repos but I will @cboettig and @noamross make that call.  If trailing whitespace is your overriding concern, is there possibly a toggle in your editor?  Or could you use a different editor for these files?

damacus:
I'm happy to break this PR up into separate sections. The build scripts are what I'm really interested in fixing long term. And there are a few typos and style thinsg I'd like to change to make it easier to read. 

If you have specific guidance on contributing to this repo, it's quite helpful to have a section in the main README or a CONTRIBUTING.md for contributing.

cboettig:
Thanks @damacus, these are all very good points and we appreciate the contribution!  

Actually if you could break the PR up between whitespace changes and everything else, that would be super handy for reviewing it.

damacus:
Brilliant, will do!

damacus:
@cboettig that should be good to go. let me know if you spot any non-white space changes.

cboettig:
Sweet, appreciate it. Looks good.

 Definitely nice to get these scripts clean as it should make them easier for the community to modify and contribute back to.

PR #117, opened 2021-02-12
Standardise on using bash in scripts

damacus:
We only use S6's with-contenv/bash where we need the system variables at
runtime.

This makes a little less surprising when you suddenly don't have access to some of the bash-isms in scripts.

### DIFF
--- a/scripts/bh-proj.sh
+++ b/scripts/bh-proj.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 set -eu
 
 mkdir proj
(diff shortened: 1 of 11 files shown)


cboettig:
yeah, standardizing this is definitely a good call.  With Dockerfiles speaking `sh` instead of `bash` by default, we initially went with that, but discovered that access to those bash-isms is handy.  I think it makes sense to stick with bash then instead of trying to keep everything `sh` compatible.

eddelbuettel:
> [Y]eah, standardizing this is definitely a good call. 

I also prefer it but I also tend to give in (when I must) to those (such as: BDR at CRAN, Alpine users with minimal `/bin/sh`, things like Debian Policy, ...) who remind us every now and then to use a minimal shell.  Full `bash` is a lot nicer.  We should clearly document that we picked it and move on and enjoy it.

PR #118, opened 2021-02-12
Use multiline environment variables to reduce the number of layers

damacus:
This isn't possible in some cases where it relies on a previous
variable, for example in the binder install.


cboettig:
Well apologies here, I believe we deprecated the use of the `partials` so that the Dockerfile construction is now specified entirely by the the JSON files in "stacks".  I believe the partials files just all need to be purged.  cc @noamross .  

Dockerfiles are constructed entirely from the stack .json files using https://github.com/rocker-org/rocker-versioned2/blob/master/make-dockerfiles.R.  I believe we decided the whisker templates weren't really adding anything, and just made extension more cumbersome.  the JSON notation we defined in the stack files however doesn't give a good mechanism to distinguish when ENV needs to be sequential and when it can be collapsed.  

We did look around for a more standardized mechanism for templating, but didn't find anything convincing.  We used to be much more sensitive to number of layers back in the early Docker days when we found it easy to hit the cap, but these days the cap is so much higher -- are you seeing any issues due the excess layers or is this more aesthetic in nature?

damacus:
If we're getting rid of them, then I'm happy to purge 👍🏼 It's a little bit quicker to pull down 30 layers over 60. [There's an option to squash layers when building too](https://docs.docker.com/engine/reference/commandline/build/#squash-an-images-layers---squash-experimental). But I'm not sure we want to go there

cboettig:
:+1:

PR #119, opened 2021-02-12
Add markdown and shell linting

damacus:


cboettig:
I like where this is going!  Looks like you may still  be tweaking on the last check (?)

damacus:
You might want to squash and merge this PR, given the tinkering around with the Actions right at the end 👍🏼 

I've removed the yamllint step for now. The failure isn't happening locally with the same version of yamllint.

cboettig:
done, and thanks again, really slick here!

PR #121, opened 2021-02-12
Configure RStudio Server logging

### LINKED ISSUE #115: Logging?
hi! 👋🏼 

thanks for the excellent RStudio image. I notice that we're getting no logs from S6 and stdout/stderr. This makes it really hard to debug container errors. 

The only logs I'm seeing from the container right now are the logs from the startup scripts. The rstudio-server process doesn't seem to be logging correctly.

Am I doing something wrong here?

damacus:
Resolves #115

- Configures RStudio logging via `/etc/rstudio/logging.conf`
- Stop using legacy back tick notation for bash sub-shell
- Use Bash string replace instead of echo & sed
-  Quote variables wherever possible

### DIFF
--- a/scripts/install_rstudio.sh
+++ b/scripts/install_rstudio.sh
@@ -69,7 +69,7 @@ rm rstudio-server-*-amd64.deb
 
 ## RStudio wants an /etc/R, will populate from $R_HOME/etc
 mkdir -p /etc/R
-echo "PATH=${PATH}" >> ${R_HOME}/etc/Renviron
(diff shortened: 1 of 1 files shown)


cboettig:
Wow thanks! that's very nice. Could never find the official documentation for enabling that; do you have a link handy?

Also thanks for the script tidy here, also appreciated!

damacus:
Wow, took me a while to dig through my history, it's pretty well [hidden here](https://docs.rstudio.com/ide/server-pro/latest/server-management.html)

PR #122, opened 2021-02-12
GitHub Actions: Use official Docker login action

damacus:
This fixes the credential warning when using the old method


cboettig:
Nice, thanks!

PR #130, opened 2021-02-24
Use Buildx Bake for faster build times

damacus:


damacus:
I'm interested to see if this is a thing we want here. I'm quite keen to see if we can speed up the build time and parallelise more of the builds.  

## Long term goals

- Have all the Dockerfiles build using one large docker-compose.yaml file
- Use `depends_on` to mark images that require building first
- 

### Pros

- Less files to maintain
- Parallelised builds using the "new" `buildx builder`

### Cons

- Moving how the images are built might not be necessary

cboettig:
Thanks @damacus , we're definitely willing to consider more improvements to the build process.  Faster, more parallelised builds would be nice, but we also want to maintain (or rather improve) the modularity of the build process. 

We actually do use `depends_on` in some compose files, like [core](https://github.com/rocker-org/rocker-versioned2/blob/master/compose/core-4.0.4.yml) and [cuda](https://github.com/rocker-org/rocker-versioned2/blob/master/compose/ml-cuda10.1-4.0.4.yml) stacks.  We actually intended to put everything in a single compose file initially, (at least within a fixed version, say), and I think this could be done by merging the relevant stack .json files.

  IIRC, one barrier was with the CI system -- we hit build timeouts or size limits trying to build the whole stack that way, so it was useful for some of the later images in the stack to be in separate files that didn't assume GH-Actions could build all the base images first.  (Even so GH-Actions struggles with those larger images as you probably noticed).     We could go back to DockerHub based builds, but would need CRON triggers and their system has been a bit cumbersome.

Not sure if any of that is helpful, but in meantime `buildx` seems like a reasonable step.

damacus:
That's cool. I'm going to close this out as I'm moving jobs and won't have time to focus on this. But feel free to take my code as an example 👍🏼

PR #143, opened 2021-04-15
Remove duplicate COPY section form r-ver-cuda

eitsupi:
`/rocker_scripts` already exist because these images are based on `rocker/r-ver`.
I don't think we need to copy scripts again in the Dockerfiles.


cboettig:
Good catch, thanks for changing both the stack files and Dockerfiles.

PR #145, opened 2021-04-18
Delete cache files

eitsupi:
Some installation scripts does not delete the cache files, so leaving the cache files in the images.

For example, `rocker/rstudio` has apt cache files from `install_s6init.sh` `install_pandoc.sh`.

```bash
$ docker run --rm -it rocker/rstudio ls -lh /var/lib/apt/lists
total 27M
```

`rocker/shiny` has R package cache files from `install_shiny_server.sh`.

```bash
$ docker run --rm -it rocker/shiny ls -lh /tmp/downloaded_packages
total 49M
```

I think it's a good idea to bring Clean up section to the end of each script, like `install_shiny_server.sh`, to make the contents of the script easier to understand.

### DIFF
--- a/scripts/add_ubuntugis.sh
+++ b/scripts/add_ubuntugis.sh
@@ -14,3 +14,6 @@ apt-get update \
   wget \
   ca-certificates \
   && add-apt-repository --enable-source --yes "ppa:ubuntugis/ubuntugis-$UBUNTUGIS_VERSION"
+
+# Clean up
+rm -rf /var/lib/apt/lists/*
--- a/scripts/dev_osgeo.sh
+++ b/scripts/dev_osgeo.sh
@@ -127,3 +127,6 @@ R CMD INSTALL sf
 R CMD INSTALL lwgeom
 R CMD build --no-build-vignettes --no-manual stars
 R CMD INSTALL stars
+
+# Clean up
+rm -rf /var/lib/apt/lists/*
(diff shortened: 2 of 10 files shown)


cboettig:
Absolutely, thanks for this :pray:

eitsupi:
Thank you for merging soon.

I didn't include it in this PR, but it looks like it would be better to have a Clean up section last in other scripts as well.
For example, the `install_cuda-10.1.sh` repeats `apt-get update` and `rm -rf /var/lib/apt/lists/*` now, so we can expect a slight reduction in build time by organizing them.
May I submit a PR to organize the Clean up section including these changes?

cboettig:
@eitsupi Thanks -- the cuda files in particular are adapted from the corresponding NVIDIA dockerfiles.  Yeah it doesn't make sense that we repeatedly update and purge cache now that those lines are in a shell script that gets run in a single layer in the Dockerfile, but we may at some point actually want to break the cuda installation into something more modular anyway, e.g. separating out the runtime installation from the dev installation etc.  

In general I think it makes sense to either put all clean at the end, or otherwise just have it after every apt-get update block (which makes it harder to lose this step when scripts are refactored).  I don't think it's a very large contributor to build times either way.

eitsupi:
@cboettig
> it makes sense to either put all clean at the end, or otherwise just have it after every apt-get update block (which makes it harder to lose this step when scripts are refactored).

I agree it's easier to see a series of blocks starting with `apt-update`.
Thank you for your polite reply !

eddelbuettel:
(Also, we learned along the way that layers really are distinct and a 'later' `rm` cannot remove cache content from an earlier one so on the margin 'more' `rm` is better for clarity even if it seems extra verbose.)

PR #146, opened 2021-04-18
add pagedown package and Chrome in verse

RLesur:
This pull request proposes: 
- to install the [`pagedown`](https://cran.r-project.org/package=pagedown) package in `rocker/verse`
- to add an install script for Chrome which is required to build PDF files using `pagedown`

Chrome is installed through an independent script because I think it could be usefull for other purposes: for instance, the [`ẁebshot`](https://cran.r-project.org/package=webshot) package will be superseded by the [`webshot2`](https://github.com/rstudio/webshot2) package which requires Chrome.

This pull request does not propose to install Chromium for the following reason: since Ubuntu 19.04, Chromium is only available as a Snap package and `snapd` does not work in containers. There is no debian package for Chromium. IMHO, the most simple alternative is to use Google Chrome.

### DIFF
--- a/scripts/install_chrome.sh
+++ b/scripts/install_chrome.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+set -e
+
+# install Google Chrome instead of Chromium because Chromium is only available as a Snap package
+curl -LO https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
+
+apt-get update && apt-get install -y --no-install-recommends \
+        ./google-chrome-stable_current_amd64.deb
+
+rm google-chrome-stable_current_amd64.deb
--- a/scripts/install_verse.sh
+++ b/scripts/install_verse.sh
@@ -70,9 +70,12 @@ wget "https://travis-bin.yihui.name/texlive-local.deb" \
 ## Install texlive
 /rocker_scripts/install_texlive.sh
 
+## Install Chrome for pagedown support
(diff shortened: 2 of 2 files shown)


cboettig:
Thanks! adding `pagedown` to verse seems reasonable.  

`pagedown` doesn't need Chrome though, amiwrong?  I'm really hesitant to start pulling that in by default for several reasons: 

- We've tried to be pretty conservative about adding system dependencies outside of the standard ubuntu repositories because it can get really tricky to maintain in working order, particularly on the 'frozen' images
- Reluctant to package the non-open-source version of the browser (though I hear you on the issue of snaps)
- Reluctant to package trademarked version Chrome, need to check terms first (even with open source stuff one can still need permission to re-distribute trademarked content, as we did in order to include RStudio Server )
- Feels kinda heavy, even for `verse`?

I know with stuff like `webshot2` and friends there's a lot that can be done with chrome bindings, and sad that there's not a headless browser or something in the standard Ubuntu/debian repositories that can fill this niche.  Anyway would welcome more discussion on this one from anyone interested.

RLesur:
Thanks, @cboettig for your comments!

You're 💯 right, redistributing Chrome is a very bad idea.  
I should have thought about that before opening this pull request.
Sorry, it seems that I've wasted your time.

For now, I'm closing this PR. IMO, we should find another mean in {pagedown}, maybe inspired by {decapitated}.

Thanks a lot for your work on this project, I really appreciate!

PR #153, opened 2021-04-25
Update the `install_R_ppa.sh` script

eitsupi:
I was curious, so I tried a multi-architecture build of amd64 and arm64 with the same contents as `rocker/r-ver` using the script in this repository, but it took about 2 hours to build R4.0.5 alone.
https://github.com/eitsupi/r-ver

So I decided to switch to installing from PPA to save time, and noticed that `install_R_ppa.sh` had not been updated for a long time compared to `install_R.sh`.
I have tried to reflect the changes made to `install_R.sh`, so please merge it if you want. If you don't need it, please close this PR.

There are many things I don't understand about this script, so if you can advise me I will fix it.

~~By the way, I was not able to build a arm64 image even from Ubuntu 18.04 with this modified script, so I ended up reverting to building R from the source code.......~~
(Update) I completely misunderstood, I get that there is no arm version build in the PPA.

### DIFF
--- a/scripts/install_R_ppa.sh
+++ b/scripts/install_R_ppa.sh
@@ -1,23 +1,57 @@
 #!/bin/bash
 set -e
 
-UBUNTU_VERSION=${UBUNTU_VERSION:-focal}
+apt-get update && apt-get -y install lsb-release
+
+UBUNTU_VERSION=${UBUNTU_VERSION:-`lsb_release -sc`}
 CRAN_LINUX_VERSION=${CRAN_LINUX_VERSION:-cran40}
 LANG=${LANG:-en_US.UTF-8}
 LC_ALL=${LC_ALL:-en_US.UTF-8}
-DEBIAN_FRONTEND=noninteractive
+CRAN=${CRAN:-https://cran.r-project.org}
+
+##  mechanism to force source installs if we're using RSPM
+CRAN_SOURCE=${CRAN/"__linux__/$UBUNTU_VERSION/"/""}
+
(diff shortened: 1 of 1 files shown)


cboettig:
thanks, I'll take a look.  as you saw, that PPA script is really out of date. Now that we have RSPM binary builds in the version stack, I think it might be cleaner if we just dropped the PPA script?  

yeah, obviously ARM is a whole different challenge with binaries.  I don't have any ARM machines handy for testing so haven't really explored that yet on the versioned stack.  Some issues come up about that occasionally, particularly for rstudio-server.   Dirk has added support for a variety of different architectures in the official r-base channel.

eddelbuettel:
> added support for a variety of different architectures in the official r-base channel

A good part of our mojo comes from contributed binaries: the PPA via 'bspm', the RSPM install-step binaries _sans_ depends, and so on.  None of that exists for ARM so all we have is the usual Debian binaries: a much smaller subset.

eitsupi:
The PPA script isn't currently in use, so I think it's okay to remove it, or move to `depricated scripts`?
It might be a useful way to easily install R on Ubuntu as I had to use this time, but it seems that few people want to use it that way.

@eddelbuettel
I recently learned that you have maintained the Debian binaries for many years. Thanks a lot!

eitsupi:
I close this because this PR is outdated.
Recently r-hub has started distributing R binaries for Ubuntu on GitHub, so if we are building R Docker images for arm64 Ubuntu, we may use this to build in a short time.
https://github.com/r-hub/R/releases

PR #156, opened 2021-05-06
Add installation of libcurl to `install_tidyverse.sh`

eitsupi:
Attempts to install the `tidyverse` package from source by using the `install_tidyverse.sh` script will fail due to the lack of `libcurl4-openssl-dev`.
`libcurl4-openssl-dev` is a dependency of [the `curl` package](https://github.com/jeroen/curl#installation).

We can confirm that the `curl` package cannot be installed on `rocker/r-ver:devel` without `libcurl4-openssl-dev` with the following command.

```shell
$ docker run --rm -it rocker/r-ver:devel install2.r curl
```

tested image digest: `rocker/r-ver@sha256:300b6a99e37594b4829787f311f68769491e56acbe40c6225f851503d15942a2`

NOTE
Since `rocker/rstudio:devel` has `libcurl4-openssl-dev` installed by the `install_rstudio.sh`, the `curl` package installation will not fail.
Similarly, `rocker/shiny:devel` has `libcurl4-gnutls-dev` installed by `install_shiny_server.sh`, so it doesn't seem to fail to install the `curl` package.

### DIFF
--- a/scripts/install_tidyverse.sh
+++ b/scripts/install_tidyverse.sh
@@ -14,6 +14,7 @@ apt-get update -qq && apt-get -y --no-install-recommends install \
     libsqlite3-dev \
     libssh2-1-dev \
     libxtst6 \
+    libcurl4-openssl-dev \
     unixodbc-dev && \
   rm -rf /var/lib/apt/lists/*
 


eddelbuettel:
Good catch, it is a known issue with RSPM binaries (that lack actual dependencies) I also ran into. I have a (small, experimental) helper package that  resolves all (system) shared libraries from all the per-R-package shared libraries ([chshli](https://github.com/eddelbuettel/chshli) on a Debian or Ubuntu system, but I don't use it much as I work more with 'BSPM' than 'RSPM'.

eitsupi:
@eddelbuettel Thanks for your reply.
I didn't know that R don't check the dependencies when installing from RSPM, so I got new knowledge.
(To confirm, when I ran `install2.r --error --skipinstalled --repos ${CRAN} --ncpus -1 tidyverse` on `rocker/r-ver:latest`, the installation succeeded, but `library(tidyverse)` failed)

However, in this case (the `curl` package), if we install the binary from RSPM, the sample code written in the README.md works without `libcurl4-openssl-dev`, so it seems that `libcurl4-openssl-dev` is only required when installing from source.

eddelbuettel:
@eitsupi that "comes with the territory" -- the RSPM files are narrow tar.gz archives of what R installed.  That gives it a strong implicit assumption that the target machines looks like the build machine.  Works well enough for static linking (as in Windoze) but can go belly up unless care is taken with shared libraries.  It's still a very decent initiative, but one has to be aware of the limitations.

I wrote a few blog posts about all that when I poked around, about a year ago, Go to http://dirk.eddelbuettel.com/blog/code/r4/ and scroll down a little til you hit the few RSPM posts.

eitsupi:
@eddelbuettel Thank you for the detailed explanation on your blog.
Fortunately, I have not encountered this drawback of installing packages from RSPM, because the dependencies of the packages I usually use were already installed in the `rocker/tidyverse` that I often use.
BSPM looks really good too!

Issue #151, opened 2021-04-22
R in rocker/r-ver does not respect the container timezone setting.


eitsupi:
Images build from this repository ignore the OS environment variable `TZ`.
That behavior is differently from other rocker images.

For example, since I am in Japan, I sometimes set Japan time.
```shell
$ docker run --rm -it -e TZ=Asia/Tokyo rocker/r-base Rscript -e "Sys.time()"
[1] "2021-04-22 19:11:18 JST"
$ docker run --rm -it -e TZ=Asia/Tokyo rocker/r-ver:3.6.3 Rscript -e "Sys.time()"
[1] "2021-04-22 19:12:26 JST"
$ docker run --rm -it -e TZ=Asia/Tokyo rocker/r-ver:4.0.5 Rscript -e "Sys.time()"
[1] "2021-04-22 10:13:25 UTC"
```

This is due to the following line added by 6e988b319effd6eb143c650c9e023e8a8c4657f6
https://github.com/rocker-org/rocker-versioned2/blob/391713d5d3797890d2f85c600e5520f05c5f715b/scripts/install_R.sh#L144

Why is it set like this?
Is it related to this issue? https://github.com/rocker-org/rocker-versioned/issues/89

I don't know the need to write `TZ` to `Renviron` file, but how about writing `TZ` to `Renviron` when the container is started by setting [ENTRYPOINT](https://docs.docker.com/engine/reference/builder/#entrypoint)?

cboettig:
@eitsupi Yes, it's related to https://github.com/rocker-org/rocker-versioned/issues/89 plus the fact that RStudio doesn't inherit environmental variables from system, only from Renviron.  To solve #89 we need TZ in the system Renviron when RStudio starts.  

RStudio's choice not to respect system environmental variables is a constant nuisance for rocker users and a source of innumerable issues in these threads, but it's still not clear to me what the most consistent strategy would be.  In your second two examples above, it you switched from r-ver to the RStudio interface and ran the same script, can you predict what you would get for TZ in each case?  Which one is what you would expect?  (current images will give the same TZ either way, the older 3.6.3 will give you a different TZ.  


Yes, we could restore the old behavior by insisting that env vars needed by RStudio are only run when RStudio is launched.  The natural way to do this I think would be in one of the setup scripts run by `init`, i.e. as part of user_config.sh in the install_rstudio.sh configuration.  


[...]

PR #157, opened 2021-05-12
Create install_julia.sh

### LINKED ISSUE #128: install_julia script
Given the growing integration of R and Julia, an `install_julia.sh` script could be a welcome addition to the stack.

I have an example [here](https://github.com/grantmcdermott/sceptic-priors/blob/master/Dockerfile#L22), which rather lazily leverages @abelsiqueira handy [JILL](https://github.com/abelsiqueira/jill) script. (OTOH I've also seen Julia devs recommend this as a preferred option for installing Julia on Linux 🤷) 

Of course, one could always take the main ingredients out of Abel's script and include them directly in a dedicated `install_julia.sh` file. Speaking of which...

While the Julia-R interoperability generally works great (esp. from the Julia side), there is a [known issue](https://juliainterop.github.io/RCall.jl/stable/installation/#Failure-on-recent-Linux-distributions-1) associated with recent Linux builds. Adding the following to my Dockerfile solved the problem for me:

```sh
## Extra steps needed for RCall library (Julia)
ENV LD_LIBRARY_PATH="/usr/local/lib/R/lib"
RUN cp /usr/lib/x86_64-linux-gnu/libstdc++.so.6 /opt/julias/julia-1.5.0/lib/julia
```

Thanks for considering!

grantmcdermott:
Would fix #128

I expect you'll want to change certain things about the script, but I just tested locally and it seems to work fine for me. Happy to continue the discussion here.

### DIFF
--- a/scripts/install_julia.sh
+++ b/scripts/install_julia.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+set -e
+
+# Latest Julia version (this needs to be manually updated, periodically)
+JULIA_LATEST=1.6-latest
+
+JULIA_VERSION=${1:-${JULIA_VERSION:-latest}}
+
+if [ "$JULIA_VERSION" = "latest" ]; then
+    JULIA_VERSION=$JULIA_LATEST
+fi
+JULIA_MAJOR_VERSION=${JULIA_VERSION:0:3}
+
+# Download and extract Julia. We'll use a common /opt/julia/ location regardless of version 
+wget https://julialang-s3.julialang.org/bin/linux/x64/$JULIA_MAJOR_VERSION/julia-$JULIA_VERSION-linux-x86_64.tar.gz -O /opt/julia.tar.gz
+mkdir /opt/julia
+tar zxvf /opt/julia.tar.gz -C /opt/julia --strip-components 1
+rm -rf /opt/julia.tar.gz
+
+# Add to path
+echo 'export PATH="/opt/julia/bin:$PATH"' >> .bashrc
+source .bashrc
+
+# Add RCall package for interoperability from Julia
+julia -e 'using Pkg; Pkg.add("RCall")'
+
+# Add R packages for interoperability the other way around
+#Rscript -e 'install.packages(c("JuliaCall", "JuliaConnectoR")); JuliaCall::julia_setup()'
+install2.r --error --skipinstalled JuliaCall JuliaConnectoR
+Rscript -e 'JuliaCall::julia_setup()'
+
+# Make sure we start in bash
+CMD ["bash"]


scripts/install_julia.sh
```diff
@@ -0,0 +1,33 @@
...
+
+# Download and extract Julia. We'll use a common /opt/julia/ location regardless of version 
+wget https://julialang-s3.julialang.org/bin/linux/x64/$JULIA_MAJOR_VERSION/julia-$JULIA_VERSION-linux-x86_64.tar.gz -O /opt/julia.tar.gz
+mkdir /opt/julia
+tar zxvf /opt/julia.tar.gz -C /opt/julia --strip-components 1
+rm -rf /opt/julia.tar.gz
+
+# Add to path
+echo 'export PATH="/opt/julia/bin:$PATH"' >> .bashrc
+source .bashrc
```
eitsupi, line 22: Checking other scripts, I think there is room for modification in the way the PATH is set.

https://github.com/rocker-org/rocker-versioned2/blob/3b41d37f6d4c008b26f5cc8c0083f960b86be6cc/scripts/install_pyenv.sh#L22-L24

How to set environment variables is currently being discussed in https://github.com/rocker-org/rocker-versioned2/issues/151#issuecomment-834574069, but the above description seems to be the standard at this time.

I am not an expert in these scripts or PATH settings, so if this comment is incorrect, please feel free to ignore it.

eitsupi:
I think we may extract the latest stable version of Julia from [the json file which Julia versions' information is recorded](https://julialang-s3.julialang.org/bin/versions.json).

If we are allowed to use the `jsonlite` and the `purrr` packages and pipe operator in R, we can get `1.6.1` like this.

```r
library(jsonlite)
library(purrr)
library(dplyr)

js <- jsonlite::read_json("https://julialang-s3.julialang.org/bin/versions.json")

versions <- names(js)

is_stable <- purrr::map(js, ~ .x$stable == TRUE) %>%
  purrr::as_vector()

versions[is_stable] %>%
  sort(decreasing = T) %>%
  {.[1]}
```

I don't know how to do the same thing without installing any packages or tools...

eitsupi:
When we install `JuliaCall`, the `yaml` package is also installed, and we can use `yaml::read_yaml` to read json file, so it looks like we can get the latest Julia version without any additional packages.

```r
js <- yaml::read_yaml("https://julialang-s3.julialang.org/bin/versions.json")

versions <- js |>
  names()

is_stable <- Map(function(x) x$stable, js) |>
  unlist()

versions[is_stable] |>
  sort(decreasing = TRUE) |>
  (function(x) x[1])()
```

```shell
$ Rscript -e "
js <- yaml::read_yaml('https://julialang-s3.julialang.org/bin/versions.json')

versions <- js |> \
  names()

is_stable <- Map(function(x) x\$stable, js) |> \
  unlist()

versions[is_stable] |> \
  sort(decreasing = TRUE) |> \
  (function(x) x[1])() |> \
  cat()" | cat
1.6.1
```

Issue #162, opened 2021-05-22
Dockerfile update(ex. CRAN URL) improvement


eitsupi:
Currently, the configuration of the CRAN URL is done manually, and there have been several misconfigurations in the past. #127 #141
(By the way, the CRAN URL for 4.0.5 is currently set to May 19, 2021, but since 4.1.0 was released on May 18, wouldn't it be better to re-set it to May 17?)

I think the following steps to determine the CRAN URL can be automated, and GitHub Actions may be used to automatically update the URL.

1. Detect the R release date.
2. Detect the Ubuntu LTS's codename.
3. Find the CRAN URL closest to the R release date.

We can check the release date of R by referring directly to the R SVN repository with the `rversions::r_versions` function.

```shell
$ docker run --rm -it rocker/tidyverse Rscript -e "tail(rversions::r_versions())"
    version                date                nickname
118   4.0.1 2020-06-06 07:05:16          See Things Now
119   4.0.2 2020-06-22 07:05:19        Taking Off Again
120   4.0.3 2020-10-10 07:05:24 Bunny-Wunnies Freak Out
121   4.0.4 2021-02-15 08:05:13       Lost Library Book
122   4.0.5 2021-03-31 07:05:15         Shake and Throw
123   4.1.0 2021-05-18 07:05:22         Camp Pontanezen
```

The release dates and codenames of Ubuntu can be found in the `/usr/share/distro-info/ubuntu.csv` file included with Ubuntu.

```shell
$ docker run --rm -it rocker/tidyverse Rscript -e "tail(read.csv('/usr/share/distro-info/ubuntu.csv'))"
     version          codename  series    created    release        eol
29     18.10 Cosmic Cuttlefish  cosmic 2018-04-26 2018-10-18 2019-07-18
30     19.04       Disco Dingo   disco 2018-10-18 2019-04-18 2020-01-18
31     19.10       Eoan Ermine    eoan 2019-04-18 2019-10-17 2020-07-17
32 20.04 LTS       Focal Fossa   focal 2019-10-17 2020-04-23 2025-04-23
33     20.10    Groovy Gorilla  groovy 2020-04-23 2020-10-22 2021-07-22
34     21.04     Hirsute Hippo hirsute 2020-10-22 2021-04-22 2022-01-22
   eol.server    eol.esm
29
30
31
32 2025-04-23 2030-04-23
33
34
```

[...]

Issue #125, opened 2021-02-20
rocker/r-ver:latest seems not to be auto-updated.


eitsupi:
The rocker/r-ver:4.0.3 was updated but rocker/r-ver:latest image was not updated 4 days ago, is it correct behavior ?

![image](https://user-images.githubusercontent.com/50911393/108586771-df611380-7393-11eb-8b4c-cf5c9467a74d.png)

eddelbuettel:
Updates can happen on automated (i.e. `cron` job) pushes.  New entries may have to made manually.  And the project is a volunteer project ...

eitsupi:
@eddelbuettel Excuse me for not having enough words. What I wanted to say was that "4.0.3" was auto-updated but "latest" didn't seem to be auto-updated.

eddelbuettel:
It may have failed. One would have to look at the build log (which, IIRC, are not 'globally open').  Time permitting, I am sure Carl or Noam will look into.  I updated things over on my side and we have `rocker/r-base`, and from it the official `r-base`, for 4.0.4 both since Tuesday.

eitsupi:
@eddelbuettel Thank you for your response. Also, thank you for your continuous image updates.

I didn't understand that build failures were common or not, so I created this issue.
I hope it's not a big problem.

cboettig:
yeah apologies, should have posted yesterday.  Eventually these will be fully automated again, some of the stack is too large to build on GitHub Actions reliably.

PR #164, opened 2021-05-30
Add scripts and workflow definitions for automatically update core dockerfiles

eitsupi:
# Overview

A part of #162

The following steps should allow you to generate Dockerfiles for `rocker/r-ver`, `rocker/rstudio`, `rocker/tidyverse`, and `rocker/verse` (only the latest two versions of R)

1. Please work in `rocker/tidyverse`.
2. Execute `install2.r --error --skipinstalled -r https://r-lib.github.io/p/pak/dev/ -n -1 pak` to install the development version of [`the pak package`](https://github.com/r-lib/pak).
3. Run `./make-stacks.R` to create the core stack files.
4. `make setup` to generate dockerfiles and compose files.

# ToDo

- [x] Automated update dockerfiles by GitHub Actions.


eitsupi:
@cboettig There appears to be an existing workflow for updating dockerfiles, but this does not appear to be working at the moment.
When I ran `make dockefiles` in my local environment, nothing happened.

https://github.com/rocker-org/rocker-versioned2/blob/4d40a1c4a01e62f9bed0294cc6229facb39e6377/.github/workflows/dockerfiles.yml#L13-L14

I'd like to add a workflow that runs the sequence of stackfiles generation, dockerfiles generation, compose generation every day and creates a PR when there is a difference. Is it better to edit the above action rather than adding a new workflow?

dockerfiles/Dockerfile_r-ver_4.1.0
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:20.04
+FROM ubuntu:focal
```
cboettig, line 1: why prefer `focal` ?  these are technically identical, it seems that 20.04 is more systematically meaningful to users not familiar with ubuntu release chain?

dockerfiles/Dockerfile_rstudio_4.1.0
```diff
@@ -6,7 +6,7 @@ LABEL org.opencontainers.image.licenses="GPL-2.0-or-later" \
       org.opencontainers.image.authors="Carl Boettiger <cboettig@ropensci.org>"
 
 ENV S6_VERSION=v2.1.0.2
-ENV RSTUDIO_VERSION=latest
+ENV RSTUDIO_VERSION=1.4.1106
```
cboettig, line 9: I think this should still be `latest`, in line with current convention (i.e. R packages are not frozen on 4.1.0 yet either, but are always the "latest" from CRAN, and 4.1.0 docker tag is currently an alias for `latest`, so it makes sense that it's versions are likewise too.  

Maybe that doesn't matter now that the script will effectively automatically update it anyhow?

dockerfiles/Dockerfile_verse_4.0.5
```diff
@@ -5,7 +5,7 @@ LABEL org.opencontainers.image.licenses="GPL-2.0-or-later" \
       org.opencontainers.image.vendor="Rocker Project" \
       org.opencontainers.image.authors="Carl Boettiger <cboettig@ropensci.org>"
 
-ENV CTAN_REPO=http://mirror.ctan.org/systems/texlive/tlnet
+ENV CTAN_REPO=http://www.texlive.info/tlnet-archive/2021/05/17/tlnet
```
cboettig, line 8: nice catch!  clearly the reason we need automation here.

stacks-templates/core.json
```diff
@@ -0,0 +1,52 @@
+{
+  "ordered": true,
+  "TAG": null,
+  "LABEL": "org.opencontainers.image.licenses=\"GPL-2.0-or-later\" \\\n      org.opencontainers.image.source=\"https://github.com/rocker-org/rocker-versioned2\" \\\n      org.opencontainers.image.vendor=\"Rocker Project\" \\\n      org.opencontainers.image.authors=\"Carl Boettiger <cboettig@ropensci.org>\"",
+  "stack": [
+    {
+      "IMAGE": "r-ver",
+      "FROM": null,
```
cboettig, line 8: Would it not be better to have the script automatically update the stack file, so these fields don't show up as `null`?  It seems less-than-ideal to me to have a config file that hides the actual values here

cboettig (commented):
This looks good, a few comments.

Re the deprecated `make dockerfiles`, I believe that task was re-named `make setup`, which just runs the existing `make-dockerfiles.R` and `write-compose.R`.  

I still need to review the new make-stacks.R more closely here, haven't tried running the script yet.  The idea is that this would update the stack files, right?  Not sure why we have some stack file fields showing up as `null` then, as noted above?

stacks-templates/core.json
eitsupi: This file is a template for `core-x.x.x.json` that is referenced from `make-stacks.R`.
The role of `make-stacks.R` is to read this template file and generate stack files with appropriate values for the `null` fields.
This means that if we want to modify the dockerfiles, we will need to modify two files: this template file and the script.
Currently it is placed in a newly created directory called `. /stacks-templates/`, but I will move it if there is a better location.

dockerfiles/Dockerfile_rstudio_4.1.0
eitsupi: I also wondered which would be better, but I think I can keep up with version updates as it is, because if I use GitHub Actions to detect daily updates, you can reflect version upgrades immediately (the things you have to is only approve bot's PR).
Since the `docker build` action is triggered when the Dockerfile is updated, writing version number approach may be a more appropriate and convenient way.

dockerfiles/Dockerfile_r-ver_4.1.0
eitsupi: This is due to the fact that the process of generating RSPM URLs requires codenames like `focal`, not version numbers like `20.04`.
There are no situations where `20.04` is needed except for specifying the base image, and since `focal` can be used here too, I used `focal` to simplify the process.
It is possible to include `20.04` as before, but it will increase the amount of content to be written in the script.

I didn't think it would cause that much confusion, since Docker Official images often specify the Debian version with a name like `buster`.

eitsupi:
@cboettig Thank you for your prompt review!
As I replied to your comment, I am using the approach of writing the fixed part to a template json file containing `null`, and then filling the null in the script to generate the final stack files.
We can manually edit the template file for `S6_VERSION` etc., and the stack file will be generated by the script based on the template file.

I didn't explain it in detail, but since we need to edit the template json file manually as described above, I also added `.editorconfig` file to make it easier to align the format with the automatically generated stack files.

eitsupi:
> Re the deprecated `make dockerfiles`, I believe that task was re-named `make setup`, which just runs the existing `make-dockerfiles.R` and `write-compose.R`.

I get it.
It seems to be a good idea to add the execution of `make-stacks.R` here and `make setup` by GitHub Actions.

stacks-templates/core.json
eitsupi: I'm sorry if I misinterpreted your intent.
It is also a good idea to include a default value that points to the latest version, such as `latest` or `devel`, instead of `null`.
However, if we set `null` as default values, we may notice the anomaly when `docker build` fails if the variable is not assigned correctly by the script for some reason.

eitsupi:
I completely rewrote the workflow for Dockerfile updates in 3473119.
This workflow will run `./make-stacks.R` and `make setup` to generate stack files, dockerfiles and compose files.
If there are any differences from the default branch, a pull request will be created.
If no differences occur, do nothing.

Check out the following results run on my fork.
https://github.com/eitsupi/rocker-versioned2/runs/2710066539?check_suite_focus=true
image

dockerfiles/Dockerfile_rstudio_4.1.0
eitsupi: Today, the automatic execution of the script informed me that a new version of RStudioServer was available for download. https://github.com/eitsupi/r-ver/pull/67/files

cboettig:
@eitsupi Thanks, this looks good.  apologies above, I hadn't noticed that this created a `stacks-template` in addition to the stacks files.  Is this necessary?  From a quick look, I believe you could just use the existing stack file(s) as their own template, updating the versions 'in-place'.  Maybe that seems less clean, but I'm hesitant to add yet another config file layer, we already have too many!

dockerfiles/Dockerfile_rstudio_4.1.0
cboettig: Agree, with automatic updating via the CI this sounds fine.

stacks-templates/core.json
cboettig: yeah, whoops, didn't realize this was a separate stacks-template file.  I'd rather not duplicate all of our stacks again to have separate `stacks` and `stacks-templates`. I think the current script could effectively use the current stack as the 'template' and update/overwrite the previous fields, rather than have these separate templates with `null` entries?

 Ideally we'd consolidate these so we have only one JSON file per stack (obviously will require updating the `make-dockerfiles.R` file as well.  In that case, the stack file would presumably use a JSON array for the version list, and the `make-dockerfiles.R` would have to understand the logic to make the mutliple dockerfiles.

eitsupi:
@cboettig I agree with you.
My intention was to make it easier to identify the files that need to be manually modified, but as you commented, it may be better to do that until we can finally reduce the stack file to just one file.

Rewrite the file to use one previous version (i.e. the latest version until the day before) as the template.

eitsupi:
@cboettig Thinking about it a bit more, I thought `devel`, which always points to the latest version, would be appropriate for the template.
So I checked `core-devel.json` and found `/opt/texlive/bin/x86_64-linux` in the `PATH` of verse. Is it safe to delete this?

```shell
$ diff ./stacks/core-devel.json ./stacks/core-4.1.0.json
3c3
<   "TAG": "devel",
---
>   "TAG": "4.1.0",
8c8
<       "FROM": "ubuntu:20.04",
---
>       "FROM": "ubuntu:focal",
10c10
<         "R_VERSION": "devel",
---
>         "R_VERSION": "4.1.0",
15c15
<         "CRAN": "https://cloud.r-project.org",
---
>         "CRAN": "https://packagemanager.rstudio.com/all/__linux__/focal/latest",
24c24
<       "FROM": "rocker/r-ver:devel",
---
>       "FROM": "rocker/r-ver:4.1.0",
26,27c26,27
<         "S6_VERSION": "v1.21.7.0",
<         "RSTUDIO_VERSION": "latest",
---
>         "S6_VERSION": "v2.1.0.2",
>         "RSTUDIO_VERSION": "1.4.1106",
39c39
<       "FROM": "rocker/rstudio:devel",
---
>       "FROM": "rocker/rstudio:4.1.0",
44c44
<       "FROM": "rocker/tidyverse:devel",
---
>       "FROM": "rocker/tidyverse:4.1.0",
47c47
<         "PATH": "/opt/texlive/bin/x86_64-linux:/usr/local/texlive/bin/x86_64-linux:$PATH"
---
>         "PATH": "/usr/local/texlive/bin/x86_64-linux:$PATH"
```

I will change the `S6_VERSION` to `v2.1.0.2` and change the base image to `ubuntu:latest` which represents the latest Ubuntu LTS.

cboettig:
@eitsupi I think we need the `/opt/texlive/bin/...` thing in path since texlive is installed there? https://github.com/rocker-org/rocker-versioned2/blob/master/scripts/install_texlive.sh  It's echoed into .Renviron but the shell won't find texlive if it's not added to PATH.

eitsupi:
@cboettig I looked up verse:4.0.0 and it looks like you commented "fix path" in bd40f16262c6b7aee3c2ca89659377f6365be0d3 and removed the `/opt/texlive/bin/x86_64-linux`...

Since verse:devel has been used without any problems so far, it looks like I don't need to remove it, so I'll adapt it to the devel side like you said.

stacks-templates/core.json
eitsupi: Done.

As for the core stackfiles, we have devel, so I changed it to use that as the template.
For other types of stack files, we may need to include a process to dynamically select a template from the file name.

cboettig:
yeah good question, I think `/opt/texlive/bin` may all be symlinked to `/usr/local/bin`?  It's been a while since I've done a deep dive on the texlive setup, it always confuses me

cboettig:
Nice, this looks good to me, thanks for this excellent contribution!  Very nice to have the release process more automated.  Here we go :rocket:

At some point we should still merge things down to single stack file per stack, but makes sense to migrate in smaller steps here.
(Also at some point should clean up which `Renviron` file is used for setting env vars consistently across stacks)

eitsupi:
Thank you for merging! I'm happy to contribute.

The fact that Docker images are updated quickly and accurately is a benefit for users like me.
I would like to continue to contribute to the automation so that the burden of version updating can be reduced.

cboettig:
:+1:  stack is rebuilding on the server, but we should definitely start moving at least the core stack to build on GH-Actions.  (it's a pity that GH-Actions self-hosted runners are not secure for use on public repositories.... but can put them on cron job locally at least)

eitsupi:
@cboettig I also think that the current docker build workflow has room for improvement.
In particular, GitHubActions does not give the `latest` tag, which is inconvenient from the user's perspective.
It looks like `tag.sh` is responsible for giving the `latest` tag, but I don't understand how this can be incorporated into GitHubActions...
(At the stage of writing #125 I didn't understand about tags, but now I understand a little bit how to give `latest` tag.)

A suggestion I can make is to completely rewrite the `core.yml` using [`docker/build-push-action`](https://github.com/docker/build-push-action) to automatically give the latest version of R the `latest` tag using the build matrix generated from `make-stack.R`.
https://github.com/eitsupi/r-ver/blob/97a06cffb9b39b725486419b3028e6fd49345f80/.github/workflows/docker-build-push.yml#L85-L92

This means that you don't have to edit `core.yml` every time you upgrade R as you do now, and if you want, we can set it to build all versions from 4.0.0 to the latest at once like that.
image
(However, the number of versions is increasing, so we may need to limit the number of versions we build at some point.)

With these changes, the core stack will no longer need to be built locally by you, and can be left to GitHubActions.
Eventually, it would be nice to be able to change all buildable stacks that way within the limits of GitHubActions.

cboettig:
@eitsupi it should be possible to just call the `make` commands in the GitHub Actions thread.  I agree the makefile could be a bit clearer.  

One trick is to be sure that we apply the `latest` tag to the image built with the current version, such that if someone does `docker pull r-ver:4.1.0` and then `docker pull r-ver:latest` they don't download any new layers with the second operation since the names are pure aliases pointing to the same image hash.  I think the GitHub actions matrix you link is doing this but not 100% sure?

I do totally support having as much fully automated on GH actions as possible, but I also prefer to keep GH-Actions configs to be more vanilla, such that it is trivial to reproduce their behavior locally without relying on GH-Actions-specific infrastructure and configuration.  That makes it easier to test locally and easier to migrate when necessary....


I think the GH-Actions should not try to build all old versions.  Ideally we'd built `latest` on maybe a daily cron job, while the other images don't need to be rebuilt so regularly; every few months is fine, probably as a separate action...

eitsupi:
@cboettig Yes, complex GitHubActions make it difficult to reproduce the work locally...

I found out that the `docker buildx bake` command, which refers to `docker-bake.json` or `docker-bake.hcl`, allows we to set multiple tags in a simple definition file.
https://github.com/docker/buildx/blob/master/docs/reference/buildx_bake.md#file-definition

So instead of the current `docker-compose.yml`, it might be easiest to generate `docker-bake.json` and write the settings for the `latest` tag in it.

> I think the GitHub actions matrix you link is doing this but not 100% sure?

Setting tags using if statements is the way described in the `docker/build-push-action` documentation, and it works well.
Here are images created with this method, and you can see that the `latest` tag is given as intended.
https://github.com/users/eitsupi/packages/container/r-ver/versions

image

> I think the GH-Actions should not try to build all old versions. Ideally we'd built `latest` on maybe a daily cron job, while the other images don't need to be rebuilt so regularly; every few months is fine, probably as a separate action...

I agree. Since the base image ubuntu is updated only once every few months, there seems to be no point in frequent updates of past images.

It is easy to switch the matrix depending on the type of trigger, so for example, the latest version is updated periodically, and the older version is built only when the file is updated, which is updated when R is upgraded.
https://github.community/t/how-to-set-schedule-event-for-two-branches-and-push-pull-request-events-for-only-one/18267

It seems easy too to update the Dockerfile with base image updates by using dated tags like `ubuntu:focal-20210416` in the Dockerfile and letting dependabot monitor it, but it's not compatible with the automatic Dockerfile generation...
https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#package-ecosystem

Issue #165, opened 2021-05-31
Incorrect cookie name results in silent failure


dhirschfeld:
https://github.com/rocker-org/rocker-versioned2/blob/4d40a1c4a01e62f9bed0294cc6229facb39e6377/scripts/install_rstudio.sh#L70-L71

After installing `rstudio-server-1.4.1106-amd64.deb` the cookie filename I see is `secure-cookie-key` instead of `secret-cookie-key`
```bash
❯ ls /var/lib/rstudio-server/
drwxr-xr-x   - root           31 May 10:49 body
drwxr-xr-x   - root           31 May 10:49 conf
drwxr-xr-x   - root           31 May 10:49 proxy
.rw-r--r-- 20k rstudio-server 31 May 10:49 rstudio.sqlite
.rw-------  32 root           31 May 10:49 secure-cookie-key
```

eitsupi:
This problem was reported in #137, at which point the file name was written as `secure-cookie-key`.
In other words, the file name has not been changed, but the modification made in 415b065d45666bbcfb435268bf30b0bfc4f10c48 seems to have specified the wrong file name.

dhirschfeld:
Right, I think the TL;DR is that the change in 415b065d45666bbcfb435268bf30b0bfc4f10c48 is effectively a no-op and as such users are still exposed to the issue identified in #137.

I've updated the issue name to be more descriptive./accurate

cboettig:
@dhirschfeld right on, thanks for catching.  will fix this

cboettig:
(also still unclear to me why RStudio is creating this file on install, as you can see we do not run RStudio executable in the install so it should not be created in the first place.)

dhirschfeld:
> *also still unclear to me why RStudio is creating this file on install, as you can see we do not run RStudio executable in the install so it should not be created in the first place.*

Yeah, I think this *should* be fixed upstream - it seems to have no benefit and admits the potential of a hard to identify security issue :/

dhirschfeld:
Fixed by #166

cboettig:
Re upstream, @kevinushey just pointed out in https://github.com/rstudio/rstudio/issues/9175#issuecomment-852388099 that RStudio does indeed seem to be executed by the debian postinstall script, just as @eddelbuettel first suggested might be the case back in #137.   Anyway, #166 should take care of this for us (once images are rebuilt) and :crossed_fingers: it is also fixed upstream

PR #166, opened 2021-06-01
Remove the authentication cookie file

### LINKED ISSUE #137: Issue with RStudio user authentication cookies
There is an issue in these Docker images with the secret used for user authentication cookies not being sufficiently unique. This file:

    /var/lib/rstudio-server/secure-cookie-key

is constant, i.e. two completely separate Docker processes running `rocker/rstudio:4.0.4` have the same secret. This makes it possible for a user to obtain authentication cookies from one Docker process and use them to log in to a different RStudio running in a different Docker container.

Please see this issue for details: https://github.com/rstudio/rstudio/issues/9175

According to the information there, a fix would be to not include that file at all in the Docker image, in which case RStudio will generate a properly unique one itself upon startup. I had a brief look at the Dockerfiles and init scripts but couldn't find where this file is currently coming from. Perhaps you can help.

eitsupi:
Fixes #137 #165

```shell
$ docker run --rm -it rocker/rstudio:4.1.0 ls -l /var/lib/rstudio-server/
total 36
drwxr-xr-x 2 root           root            4096 May 23 21:57 body
drwxr-xr-x 2 root           root            4096 May 23 21:57 conf
drwxr-xr-x 2 root           root            4096 May 23 21:57 proxy
-rw-r--r-- 1 rstudio-server rstudio-server 20480 May 23 21:57 rstudio.sqlite
-rw------- 1 root           root              32 May 23 21:57 secure-cookie-key
```

### DIFF
--- a/scripts/install_rstudio.sh
+++ b/scripts/install_rstudio.sh
@@ -68,7 +68,7 @@ dpkg -i rstudio-server-*-amd64.deb
 rm rstudio-server-*-amd64.deb
 
 # https://github.com/rocker-org/rocker-versioned2/issues/137
-rm -f /var/lib/rstudio-server/secret-cookie-key
+rm -f /var/lib/rstudio-server/secure-cookie-key
 
 ## RStudio wants an /etc/R, will populate from $R_HOME/etc
 mkdir -p /etc/R


cboettig:
thanks!

PR #169, opened 2021-06-07
Update build core image action to push `latest` tagged images

eitsupi:
Suggested by https://github.com/rocker-org/rocker-versioned2/pull/164#issuecomment-854841158

Make the following updates.

1. `make-stacks.R` will generate a new json file for the build matrix.
2. The `core.yml` will read the version of R to be built from the build matrix, and if R is the latest version, it will also push images with `latest` tag.
3. Limit the conditions under which `core.yml` is executed to only when a specific file is modified.

These changes will clarify the update timing of images on DockerHub, and images with the `latest` tag will always follow the latest R version.

**Note**
1. I haven't been able to fully debug it because I can't log in to DockerHub, but I have confirmed that the `if: matrix.r_latest == true` method to determine the latest version is working. https://github.com/eitsupi/rocker-versioned2/actions/runs/912039914
image
image
image
2. In this PR, only the `core.yml` has been changed, but if there is no problem with the behavior, I would like to apply the same changes to other workflows that build other images.
3. The current build matrix includes only one version, so the latest version can be determined without `r_latest` in the build matrix, but it is included in case you need to build multiple versions at the same time in the future.

### DIFF
--- a/.github/workflows/buildmatrix/latest.json
+++ b/.github/workflows/buildmatrix/latest.json
@@ -1,0 +1,8 @@
+{
+  "include": [
+    {
+      "r_version": "4.1.0",
+      "r_latest": true
+    }
+  ]
+}
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -1,19 +1,57 @@
 name: Build & Push Core images
 
-'on': [push]
-#  release:
-#    types: [published]
+on:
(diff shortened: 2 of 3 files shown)


cboettig:
LGTM

eitsupi:
🎉It seems to be working fine.

PR #175, opened 2021-06-14
Update for automatically update shiny, geospatial, binder

eitsupi:
A part of #162

Make the same changes to shiny, geospatial, and binder that we made to core images in #164 and #169.


eitsupi:
@cboettig Currently, `rocker/shiny` and `rocker/geospatial` can only be built after core dockerfiles due to dependencies, and `rocker/binder` can only be built after geospatial.
Therefore, as discussed in #130, or as I wrote in https://github.com/rocker-org/rocker-versioned2/pull/164#issuecomment-854891895, it is preferable to build them using a single compose file.
If we let them build separately as they do now, the number of stack files will increase, and the builds in GitHubActions will have to be configured to run in order.

Is it safe to proceed with the automation with the current stack file configuration as shown in this PR?
Or do you want to do a major overhaul of the configuration of the stack and compose files first?

Also, even if we continue with the current configuration, can I consolidate the GitHubActions workflow, geospatial and binder into one workflow?

cboettig:
@eitsupi yeah we struggled with that early on as well.  Like you say, that was originally the purpose of using compose in the first place (and they could also be in the same stack file).  Ironically it was to keep build time down that we first broke up the compose files (that may have been prior to RSPM binaries though).  I suspect we could go back to one compose (and likewise condense the corresponding stack files into one stack file) now and see if it builds?  

we may need a different solution for the geospatial-dev-osgeo, which is a real bear to build, but as you see it's already a different stack file.  

My instinct is to try the stack file consolidation first, but happy to go either route if you  have a preference.

eitsupi:
Thank you for your explanation.

Checking the execution history of the most recent GitHubActions, the total build time for core, shiny, geospatial, and binder is less than an hour.
The maximum build time for GitHubActions is 6 hours, so I don't think it will time out if we build these from a single compose file now.
image

However, changing the structure of the stacks is a bit of a challenge, so I'd like to try to combine only the workflows into one first, so that we can revert if that doesn't work.

eitsupi:
I have finished editing.
If GitHubActions does not work well, you can revert to the original GitHubActions by revert 9892499.

cboettig (approved):
Nice work! :100: All looks good here.

.github/workflows/core.yml
```diff
@@ -41,17 +41,33 @@ jobs:
...
+          docker-compose -f compose/geospatial-${{ matrix.r_version }}.yml push
+          docker-compose -f compose/binder-${{ matrix.r_version }}.yml push
       - name: Push latest tagged Docker image
         if: matrix.r_latest == true
         run: |
           docker image tag rocker/r-ver:${{ matrix.r_version }} rocker/r-ver:latest
           docker image tag rocker/rstudio:${{ matrix.r_version }} rocker/rstudio:latest
           docker image tag rocker/tidyverse:${{ matrix.r_version }} rocker/tidyverse:latest
           docker image tag rocker/verse:${{ matrix.r_version }} rocker/verse:latest
+          docker image tag rocker/verse:${{ matrix.r_version }} rocker/shiny:latest
```
cboettig, line 62: looks like these tags were not updated

eitsupi: @cboettig I apologize for not noticing this mistake. Since the build won't run if you just fix this part, could you merge #176?

PR #176, opened 2021-06-15
fix latest tagged images

eitsupi:
@cboettig I'm very sorry.
The workflow worked fine, but I misspelled the image name and accidentally pushed the `rocker/verse:4.1.0` as the latest of another image in #175.
If you merge this PR, the build will be done again and the correct images will be pushed.

image
image


cboettig:
no worries, and thanks for adding the additional trigger, looks like it's rebuilding now

Issue #13, opened 2020-02-27
GitHub actions workflow


noamross:
On commit: (@noamross)
 - Re-compile Dockerfiles, commit back to repo
    - This requires R installed and the `rockerbuild` package (#12)

Nightly: (@cboettig )
 - [ ] (post-release, on master) Ping Docker Hub to build images 
 - [x] (pre-release and post-release) Build images, push to GitHub registry

On successful Dockerfile creation with commit keyword (e.g, `[build images]`) or nightly on branch:
 - Build images, push to GitHub registry (possibly with additional tag)

cboettig:
@noamross we should finish setting up these triggers with cron jobs now that things are a bit more stable?

eitsupi:
Github actions workflows seem working well now.

PR #180, opened 2021-06-20
build core devel images every day

eitsupi:
Related to #13
This is an update for scheduled build of the devel "core" images by GitHubActions, as suggested at https://github.com/rocker-org/rocker-versioned2/issues/162#issuecomment-848940332.

These images on DockerHub may not have been updated for several days, as it is currently being built irregularly by the local computer(https://github.com/rocker-org/rocker-versioned2/issues/162#issuecomment-847145824).

### DIFF
--- a/.github/workflows/devel.yml
+++ b/.github/workflows/devel.yml
@@ -1,8 +1,8 @@
 name: Build & Push Core devel images
 
-'on': [push]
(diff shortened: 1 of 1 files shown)


cboettig:
Yes!  :tada:

PR #183, opened 2021-06-22
use etc/Renviron.site and not etc/Renviron

cboettig:
As noted in #151, we are better off appending to `R_HOME/etc/Renviron.site` and not modifying the system `R_HOME/etc/Renviron` as discussed in https://support.rstudio.com/hc/en-us/articles/360047157094-Managing-R-with-Rprofile-Renviron-Rprofile-site-Renviron-site-rsession-conf-and-repos-conf

(functionally I am not sure this changes anything, as the previous version only appended to `R_HOME/etc/Renviron`, though certainly modifying or deleting any of the system `R_HOME/etc/Renviron` settings would be bad.  In any event writing to a separate file is always cleaner).  

 @eitsupi @Pit-Storm would you care to take a look over the proposed changes?

### DIFF
--- a/compose/ml-cuda10.1-4.1.0.yml
+++ b/compose/ml-cuda10.1-4.1.0.yml
@@ -1,21 +1,21 @@
 version: '3'
 services:
-  r-ver-4.1.0-cuda11.1:
-    image: rocker/r-ver:4.1.0-cuda11.1
+  r-ver-4.1.0-cuda10.1:
+    image: rocker/r-ver:4.1.0-cuda10.1
     build:
       context: ..
-      dockerfile: dockerfiles/Dockerfile_r-ver_4.1.0-cuda11.1
-  ml-4.1.0-cuda11.1:
-    image: rocker/ml:4.1.0-cuda11.1
+      dockerfile: dockerfiles/Dockerfile_r-ver_4.1.0-cuda10.1
+  ml-4.1.0-cuda10.1:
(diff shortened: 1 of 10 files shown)


eitsupi:
I found out that this modified container is on DockerHub (#185) in the morning, so I tried to use RStudio with `rocker/tidyverse` during the day. I think it was working fine.

Do you need to change the following lines as suggested in this comment https://github.com/rocker-org/rocker-versioned2/issues/151#issuecomment-834291515 ?
(I don't understand what these variablew are for...)
https://github.com/rocker-org/rocker-versioned2/blob/7317191b25a6a6e21b54253f31b90d3c832b2dda/scripts/userconf.sh#L100-L101

cboettig:
Thanks @eitsupi !  The `HTTR_*` env vars are for the `httr` package OAuth call mechanism, when it needs to open up a port to allow a user to authenticate in a separate window.  The default is localhost, 127.0.0.1, which won't work for a remotely-hosted rstudio instance.  They aren't very important though since most OAuth systems support an alternative mechanism anyhow.  

also good catch about not having the workflow push core images on branches & PRs.

Pit-Storm:
@cboettig: We worked on it parallely, but it's not bad. I did all of it in my other PR #186.
Hopefully I didn't mess everything up. I just did a find and replace and check every line for its context. Please check if it will work :-)

Issue #182, opened 2021-06-22
`ml-cuda10.1-4.1.0.json` generate `cuda11.1` definition files


eitsupi:
Unlike the file name, `ml-cuda10.1-4.1.0.json` generates Dockerfile and compose file of cuda11.1.

Because a Dockerfiles such as `Dockerfile_r-ver_4.1.0-cuda11.1` are generated by `ml-cuda11.1-4.1.0.json` later and overwrites it, the image built by GitHubActions to build the image of cuda10.1 is actually the image defined in `ml-cuda11.1-4.1.0.json`.
The contents of both are quite different.

This doesn't seem to be an intentional behavior, do you think it can be solved by auto-generating stack files ?

```shell
$ diff ./stacks/ml-cuda10.1-4.1.0.json ./stacks/ml-cuda11.1-4.1.0.json
5c5
<   "LABEL": "org.label-schema.license=\"GPL-2.0\" \\\n      org.label-schema.vcs-url=\"https://github.com/rocker-org/rocker-versioned\" \\\n      org.label-schema.vendor=\"Rocker Project\" \\\n      maintainer=\"Carl Boettiger <cboettig@ropensci.org>\"",
---
>   "LABEL": "org.opencontainers.image.licenses=\"GPL-2.0-or-later\" \\\n      org.opencontainers.image.source=\"https://github.com/rocker-org/rocker-versioned2\" \\\n      org.opencontainers.image.vendor=\"Rocker Project\" \\\n      org.opencontainers.image.authors=\"Carl Boettiger <cboettig@ropensci.org>\"",
9c9,10
<     "FROM": "rocker/r-ver:4.1.0",
---
>     "FROM": "nvidia/cuda:11.1.1-cudnn8-devel-ubuntu20.04",
>     "COPY": "scripts /rocker_scripts",
11,19c12,18
<       "CUDA_VERSION": "11.1",
<       "NCCL_VERSION": "2.7.8",
<       "NVIDIA_VISIBLE_DEVICES": "all",
<       "NVIDIA_DRIVER_CAPABILITIES": "compute,utility",
<       "NVIDIA_REQUIRE_CUDA": "cuda>=11.1 brand=tesla,driver>=418,driver<419 brand=tesla,driver>=440,driver<441 brand=tesla,driver>=450,driver<451",
< 
<       "CUDA_HOME": "/usr/local/cuda",
[...]

PR #185, opened 2021-06-23
Change the triggers for build workflows

eitsupi:
I noticed by #183 that the images was being pushed to DockerHub even when it was pushed to a branch other than the default branch because we hadn't specified the branch in the workflow's execution condition.
This is not a desirable behavior and will be fixed.

And, I thought it was undesirable to have the workflow run with such workflow changes that do not affect the contents of the container being built, so I excluded the change of `core.yml` from the trigger and allowed it to be executed manually instead.

**Note**: I think the execution condition of `ml-10.1.yml` should be success of `core.yml` because ml-cuda10.1 is originally derived from `rocker/r-ver`, but I left it as it is because of the discussion in #182.

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -2,6 +2,8 @@ name: Build & Push Core images
 
 on:
   push:
+    branches:
+      - "master"
     paths:
       - ".github/workflows/buildmatrix/**"
       - ".github/workflows/core.yml"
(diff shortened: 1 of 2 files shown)


cboettig:
> Note: I think the execution condition of ml-10.1.yml should be success of core.yml because ml-cuda10.1 is originally derived from rocker/r-ver, but I left it as it is because of the discussion in #182.
:+1: agreed

PR #186, opened 2021-06-23
Dynamic container vars and locale Settings

Pit-Storm:
You just have to set `-e LANG=de_DE.UTF-8` (or other language code) to have it present in RStudio. If you want to have all locale Settings present in an stand alone R-Session you have to set `-e LC_ALL=de_DE.UTF-8 -e LANG=de_DE.UTF-8`. The `LANG` ENV-Var generates the needed loacle files `LC_ALL` is for convenience in a custom CMD overwrite (standalone R-Session).

Bcause the Timezone Setting is much easier, in any case you can use `-e TZ=Europe/Berlin` (or other Timezone) to use it in standalone R and/or RStudio.

No need to write custom Dockerfile for localization 🎉

There is a bug, that the script says, that it is not able to set "LC_ALL" because there where no file, but this is not effecting anything, because the locale files are properly generated.... maybe this is an restart issue, because we are not restarting like it would be when we doing it in a Dockerfile before running CMD or ENTRYPOINT.

Hoping that I did everything correct.

This is my first PR ever ✌


cboettig:
Thanks, nicely done.

Note that #183 already did the `Renviron` to `Renviron.site` edit. (generally best to do separate changes as separate PRs.  doesn't matter much, but note it creates merge conflicts in a few files above).  Anyway should be straight-forward to resolve.

Looks like this still defaults to UTC if no TZ is provided, right?  

Other than the merge conflict this looks good to me. :tada:

Pit-Storm:
You are right with everything you wrote. Seperate is better and default is UTC.

default LANG is en_US.UTF-8 like it was before.

PR #194, opened 2021-07-11
build core latest daily images every day

eitsupi:
It will build images daily, including the RStudio daily build, which is currently not built automatically.

### DIFF
--- a/.github/workflows/devel.yml
+++ b/.github/workflows/devel.yml
@@ -1,4 +1,4 @@
-name: Build & Push Core devel images
+name: Build & Push R devel images and RStudio daily build images
 
 on:
   schedule:
@@ -7,13 +7,17 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        composefile: [core-devel.yml, core-latest-daily.yml]
     steps:
       - uses: actions/checkout@v2
       - uses: docker/login-action@v1
         with:
-            username: ${{ secrets.DOCKER_USER }}
-            password: ${{ secrets.DOCKER_PASSWORD }}
-      - name: Build the tagged Docker image
-        run: docker-compose -f compose/core-devel.yml build
-      - name: Push the tagged Docker image
-        run: docker-compose  -f compose/core-devel.yml push
+          username: ${{ secrets.DOCKER_USER }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+      - name: Build Docker images
+        run: docker-compose -f compose/${{ matrix.composefile }} build
+      - name: Push Docker images
+        run: docker-compose -f compose/${{ matrix.composefile }} push
--- a/compose/core-latest-daily.yml
+++ b/compose/core-latest-daily.yml
@@ -1,21 +1,21 @@
 version: '3'
 services:
-  rstudio-4.0.5-daily:
-    image: rocker/rstudio:4.0.5-daily
+  rstudio-latest-daily:
+    image: rocker/rstudio:latest-daily
     build:
       context: ..
-      dockerfile: dockerfiles/Dockerfile_rstudio_4.0.5-daily
-  tidyverse-4.0.5-daily:
-    image: rocker/tidyverse:4.0.5-daily
+      dockerfile: dockerfiles/Dockerfile_rstudio_latest-daily
+  tidyverse-latest-daily:
+    image: rocker/tidyverse:latest-daily
     depends_on:
-    - rstudio-4.0.5-daily
+    - rstudio-latest-daily
     build:
       context: ..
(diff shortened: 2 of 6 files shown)


eitsupi:
Currently, the images built by `core-latest-daily.yml` are `rocker/rstudio:4.0.5-daily`, `rocker/tidyverse:4.0.5-daily`, and `rocker/verse:4.0.5-daily`.
The current `rocker/r-ver:latest` is the same as `rocker/r-ver:4.1.0`, so the tag with "4.0.5" on the built images are inappropriate.

To solve this, it is a good idea to make one of the following changes before setting up this automatic build workflow.
1. Change tags to something like `rocker/rstudio:latest-daily`, `rocker/tidyverse:latest-daily`, `rocker/verse:latest-daily`
2. Set the base image to `rocker/r-ver:4.1.0` and set the tag to be automatically updated by `make-stacks.R`. The disadvantage of this method is that when a new version of R is released, the scheduled workflow may work without a new image such as `rocker/r-ver:4.1.1` on DockerHub, which may cause errors.

cboettig:
@eitsupi Thanks.  Yes, I think we either want to deprecate or rework the `latest-daily` tags.  The intent was to provide rstudio:daily, which should now be possible with something like (not tested):

```
FROM rocker/rstudio:latest
RUN RSTUDIO_VERSION="daily" /rocker_scripts/install_rstudio.sh
```

rather than the existing build setup...  Meanwhile, making the core `latest` tag stack build daily seems like a very good idea, since those images are used more regularly anyway....

eitsupi:
It's definitely great to be able to easily try out the development version of RStudio, and I thought it should be built daily.

> The intent was to provide rstudio:daily, which should now be possible with something like (not tested):
> 
> ```
> FROM rocker/rstudio:latest
> RUN RSTUDIO_VERSION="daily" /rocker_scripts/install_rstudio.sh
> ```
> 
> rather than the existing build setup...

If we use `rocker/rstudio:latest` as the base image, the size should increase due to more layers, so I think it makes more sense to use `rocker/r-ver:latest` as the base image as it is now.

> Meanwhile, making the core `latest` tag stack build daily seems like a very good idea, since those images are used more regularly anyway....

There are some projects that use dependabot and other settings to monitor the digest and receive a PR every time the image is updated, so I think it is better not to update `rocker/r-ver:latest` etc. every day.

cboettig:
Thanks @eitsupi for the explanations, that all sounds good to me.  

This looks good.  Only other question: why do this in the `devel.yml` Action, shouldn't it be a separate action for "daily'?

eitsupi:
@cboettig Since RStudio's daily build is a development version of RStudio.
And, I think it would be easier to standardize the workflow into one workflow rather than separate workflows, because it is executed only once a day and all the files are the same except for the compose file used for the build.
Of course, the workflow name that is currently "Build & Push Core devel images" and the file name `deve.yml` can be changed to a more appropriate name.

Now that I think about it again, since there is no dependency between core-devel.yml and core-latest-daily.yml, how about using a build matrix to build them in parallel as follows?

```yml
jobs:
  build:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        compose: [core-devel, core-latest-daily]
    steps:
      - uses: actions/checkout@v2
      - uses: docker/login-action@v1
        with:
            username: ${{ secrets.DOCKER_USER }}
            password: ${{ secrets.DOCKER_PASSWORD }}
      - name: Build Docker images
        run: docker-compose -f compose/${{ matrix.compose }}.yml build
      - name: Push Docker images
        run: docker-compose  -f compose/${{ matrix.compose }}.yml push
```

Thanks!

cboettig:
nice.  yeah, the parallel build matrix sounds good to me!

eitsupi:
I manually updated `stacks/core-latest-daily.json` by copying the contents from `stacks/core-devel.json`.(I'm sorry I've force-pushed many times!)

Of course, we can automate the copying with a script such as `make-stacks.R`, but I haven't updated the script at this stage because it will conflict with #192.

If you have no problem with the current state, where updates are not automated, please merge. I would like to update the script with another PR after #192 merged.
Or, if the review of #192 will be finished first and merged, then I would like to update the script in this PR as well.

eitsupi:
🎉
image

image

But it didn't work with error...

```
2021-07-18T02:02:38.082768Z [rserver] ERROR database error 7 (sqlite3_statement_backend::prepare: no such table: revoked_cookie) [description: Could not delete expired revoked cookies from the database, description: Could not read revoked cookies from the database]; OCCURRED AT virtual rstudio::core::Error rstudio::core::database::Connection::execute(rstudio::core::database::Query&, bool*) src/cpp/core/Database.cpp:496; LOGGED FROM: int main(int, char* const*) src/cpp/server/ServerMain.cpp:783
```

Issue #144, opened 2021-04-16
`rocker/rstudio` for apple silicon linux/arm64/v8 platform


dvillaj:
Hi,

I am experiencing the following error when I try to run the container on a Macbook Pro M1 (MacOS 11.2.3):

WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested

![Screenshot 2021-04-16 at 05 20 29](https://user-images.githubusercontent.com/8428613/114966729-78cc1380-9e73-11eb-892e-ad8064091f56.png)

When I try to log-in in RStudio I get the following error: Error: Incorrect or invalid username/password

![Screenshot 2021-04-16 at 05 32 38](https://user-images.githubusercontent.com/8428613/114967673-2a1f7900-9e75-11eb-8012-fee17e1e82c2.png).

I have tried with the latest version of Docker Desktop (3.3.1) and disabling/enabled experimental features.

style206:
I have the same problem :(
I hope that the update supporting M1 Mac.

eitsupi:
See this issue for arm64 build of RStudio.
https://github.com/rstudio/rstudio/issues/8809

In addition, even if the above issue is resolved and RStudio for arm can be installed, RStudio Package Manager (RSPM) does not provide R binary packages for arm64, so Docker images like `rocker/tidyverse` will take a very long time to build...

eitsupi:
The upstream issue https://github.com/rstudio/rstudio/issues/8809 appears to have been registered for the 2022-02 (or 2022-03) milestone!

echuber2:
As a stopgap measure, does the amd64 version work (maybe slowly) [with Rosetta 2 emulation enabled](https://docs.docker.com/desktop/mac/apple-silicon/)? We have a version of the rocker image here that has auth turned off for limited usage; reportedly, some people are able to run it on M1 Macs: https://github.com/PrairieLearn/PrairieLearn/tree/master/workspaces/rstudio

eitsupi:
It appears that the arm64 daily build for jammy is now available.

https://rstudio.org/download/latest/daily/server/jammy/rstudio-server-latest-arm64.deb

I believe this can be installed with the `install_rstudio.sh` script, but so far there is no jammy-based, arm64-architecture `rocker/r-ver` here, so I have not tried it. (`rocker/r-ver:devel` is already jammy but amd64 only)
[...]

PR #195, opened 2021-07-11
Update some scripts to install R packages from source, on arm64 images

eitsupi:
Related to #144

The RSPM CRAN mirrors does not provide arm64 binary packages for ubuntu at this time.
So, in this PR, I will modify the rocker script so that the R package will do source installations in the arm64 images.

- `install_R.sh`: Change the default repository to be written to `Rprofile.site` to source in arm64 images.
- `install_tidyverse.sh` and `install_verse.sh`: Remove the reference to the CRAN env var from the options of the `install2.r` command for to use the default repository. (In other scripts originally did not reference the CRAN env var from the `install2.r` `-r` option.)

With these fixes, I have run `install_R.sh` on the `ubuntu:forcal` arm64 image, and `install_tidyverse.sh` on it, and confirmed that I can use R normally.

### DIFF
--- a/scripts/install_R.sh
+++ b/scripts/install_R.sh
@@ -11,6 +11,11 @@ CRAN=${CRAN:-https://cran.r-project.org}
 ##  mechanism to force source installs if we're using RSPM
 CRAN_SOURCE=${CRAN/"__linux__/$UBUNTU_VERSION/"/""}
 
+## source install if using RSPM and arm64 image
+if [ "$(uname -m)" = "aarch64" ]; then
+    CRAN=$CRAN_SOURCE
+fi
+
 export DEBIAN_FRONTEND=noninteractive
 
 # Set up and install R
--- a/scripts/install_tidyverse.sh
+++ b/scripts/install_tidyverse.sh
@@ -18,7 +18,7 @@ apt-get update -qq && apt-get -y --no-install-recommends install \
     unixodbc-dev && \
   rm -rf /var/lib/apt/lists/*
 
-install2.r --error --skipinstalled -r $CRAN -n $NCPUS \
+install2.r --error --skipinstalled -n $NCPUS \
     tidyverse \
     devtools \
     rmarkdown \
@@ -27,7 +27,7 @@ install2.r --error --skipinstalled -r $CRAN -n $NCPUS \
     gert
 
 ## dplyr database backends
-install2.r --error --skipinstalled -r $CRAN -n $NCPUS \
+install2.r --error --skipinstalled -n $NCPUS \
     arrow \
     dbplyr \
     DBI \
@@ -40,6 +40,6 @@ install2.r --error --skipinstalled -r $CRAN -n $NCPUS \
     fst
 
 ## a bridge to far? -- brings in another 60 packages
-# install2.r --error --skipinstalled -r $CRAN -n $NCPUS tidymodels
+# install2.r --error --skipinstalled -n $NCPUS tidymodels
 
  rm -rf /tmp/downloaded_packages
--- a/scripts/install_verse.sh
+++ b/scripts/install_verse.sh
@@ -49,7 +49,7 @@ fi
 # So: we can build the redland package bindings and then swap back to libcurl-openssl-dev... (ick)
 # explicitly install runtime library sub-deps of librdf0-dev so they are not auto-removed.
 apt-get install -y librdf0-dev
-install2.r --error --skipinstalled -r redland
+install2.r --error --skipinstalled redland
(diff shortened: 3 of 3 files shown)


eddelbuettel:
Out of curiousity do you know if anybody has said anything about such arm64 binaries?

eitsupi:
> Out of curiousity do you know if anybody has said anything about such arm64 binaries?

Sorry, I don't know. I looked into the architectures supported by RSPM and couldn't find any mention of arm64.

As you told me in https://github.com/rocker-org/rocker-versioned2/pull/153#issuecomment-826364061, there is currently no RStudio or even R binary in arm64 Ubuntu, so I'm convinced that's why RSPM doesn't provide arm64 binaries.

cboettig:
@eitsupi nice work, this approach makes sense.  Thanks for testing; I don't have access to any arm64 architectures and am not really up to speed on the details of multi-arch Docker -- is there a convenient way to have these images pre-build the arm64 versions, e.g. an arm64-based GH-Actions runner?

eitsupi:
@cboettig Multi-arch images can be built using the `docker buildx build` command to emulate multiple architectures on a single machine. You should be able to use it by setting it to `"experimental": "enabled"` in `~/.docker/config.json`.
https://docs.docker.com/buildx/working-with-buildx/#build-multi-platform-images

If you want to build a multi-arch image with GitHubActions, it is easy to use `docker/build-push-action`. The `docker buildx build` command is used in this action.
https://github.com/docker/build-push-action/blob/master/docs/advanced/multi-platform.md

For example, here is a multi-arch image that I ran `install_R.sh` included in b5469fc63d96883f011c89e39874df621f63dca8 using `docker/build-push-action`.

`ghcr.io/eitsupi/r-ver@sha256:22daf53a18f0335b7a391ef1c21cc61a8a4430a6c08baeda3643f8d418e2fa0e`
image
image

As described above, it is not that difficult to build a multi-arch image nowadays, but there are two disadvantages compared to the current amd64-only build.

1. It takes about two hours to run `install_R.sh` on GitHub Actions, which is more than ten times slower than the 10 minutes it would take to build only amd64.
2. It is not possible to build multiple images with dependencies with a single command like `docker-compose build`. This may be resolved in the future with improvements to the `docker buildx bake` command, but for now, one command per image is required to build.

I submitted this PR because I figured out how to support arm64 images, but I don't think there is much need to do multi-arch builds of `rocker/r-ver` in this repository until the arm64 build of RStudio is released.
Of course, if you do decide to build it, I'd be happy to help.

By the way, I am testing the arm64 image on Ubuntu installed on a Raspberry Pi 4😂

cboettig:
Excellent, thanks for this.  Right, certainly the builds from source are quite slow, but then that's why it would be nice to have pre-built images available to users.  Meanwhile, seems like we can go ahead an merge this anyway.

Yup, hopefully buildx will support the compose-like depends-on option which would fit better into the current workflow, though it wouldn't be too difficult to work around that with a Makefile or similar.

benz0li:
> By the way, I am testing the arm64 image on Ubuntu installed on a Raspberry Pi 4😂

@eitsupi Why not use a Mac mini M1 😉. There are also [less invasive] alternatives to GitHub for CI/CD.
ℹ️ https://benz0li.b-data.io/posts/2021/03/host-linux/arm64-gitlab-runners-on-your-mac-mini-m1/

eitsupi:
@cboettig Thank you for merging.

> Yup, hopefully buildx will support the compose-like depends-on option which would fit better into the current workflow, though it wouldn't be too difficult to work around that with a Makefile or similar.

Yes, moving to using buildx seems to be a good idea to consider. The ability to set multiple tags is also very attractive.

For example, I heard that if you build the following `docker-bake.json` with the `--push` option in the `docker buildx bake` command, we can build and push an multi-arch image with 4 tags set with one command (I haven't tried it yet!)

```json
{
    "target": {
        "r-ver": {
            "context": "../",
            "dockerfile": "dockerfiles/Dockerfile_r-ver_4.1.0",
            "tags": [
                "docker.io/rocker/r-ver:4.1.0",
                "docker.io/rocker/r-ver:4.1",
                "docker.io/rocker/r-ver:4",
                "docker.io/rocker/r-ver:latest"
            ],
            "platforms": [
                "linux/amd64",
                "linux/arm64"
            ]
        }
    }
}
```

It seems to make sense to proceed with the creation of a script to generate `docker-bake.json` from the stack files, and the Makefile and other files to repeat `bake`.

@benz0li Thanks for the information! The performance of the M1 chip is really amazing.

PR #196, opened 2021-07-12
Set the rocker/rstudio default non-root user's login shell to bash

eitsupi:
Currently, the shell for the `rocker/rstudio` default non-root user `rstudio` is bash on RStudio, but `/bin/sh` when logged in outside RStudio.
For many users, bash may be preferred as the login shell.

In addition, I changed the default user's home directory creation to be done with the `-m` option of the `useradd` command.


cboettig:
Yeah, I think this makes sense.  Looks good to me!

PR #198, opened 2021-07-20
Add a script for generate docker-bake.json

eitsupi:
Suggested by #195

To be able to use `docker buildx bake` to build container images, I add a script `make-bakejson.R` that generates `docker-bake.json` from the stack files.
Unlike `docker-compose.yml`, `docker-bake.json` allows you to configure multiple tags and multiple platforms for a single target, allowing you to run multi-architecture builds with a simple command.
(At the moment, neither `tags` nor `platforms` exist in the stack files, so multiple tags and platforms will not be set in `docker-bake.json`.)

To build `rocker/rstudio:4.1.0` using `bakefiles/core-4.1.0.docker-bake.json`, generated from `stacks/core-4.1.0.json`, execute the following command:

```shell
$ docker buildx bake -f ./bakefiles/core-4.1.0.docker-bake.json rstudio
```

Note that unlike the `docker-compose build` command, the `buildx bake` command builds all targets in parallel, so we need to build images one by one to build images with dependencies.

### DIFF
--- a/bakefiles/README.md
+++ b/bakefiles/README.md
@@ -1,0 +1,3 @@
+# Build configuration files for the `docker buildx bake` command.
+
+:warning: JSON files in this directory are generated from the stack files by [`make-bakejson.R`](../build/make-bakejson.R). **Don't edit manually.** :warning:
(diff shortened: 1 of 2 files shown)


cboettig:
Looks good!  Apologies for taking a while to get back to this one.

PR #202, opened 2021-07-29
Update the RStudio daily build file name regex

eitsupi:
The build of `rocker/rstudio:latest-daily` is failing for two days in a row, and the reason seems to be a change in the pattern of the file name of RStudio Server daily build.

It seems to be related to https://github.com/rstudio/rstudio/pull/9639

image
image

Current file name pattern.
image


cboettig:
Ah good catch,  I was scratching my head on that yesterday and didn't have a chance to investigate.

eddelbuettel:
Oddly enough the helper scripts I have in littler picked up the new names URLs totally seemlessly ... and I believe that is due to logic that _you_ initially put into them by parsing the webpages as a table and deriving the URL from what it is in the table :)

Issue #199, opened 2021-07-20
[Tracking] Use `buildx bake` to build images


eitsupi:
- [x] Add a script for generate `docker-bake.json` files from the stack files. (#198)
- [x] Minor modifications to the stack files and the existing scripts to support setting multi-platform and multi-tagging.
  - [x] `stacks/binder-.*json` (#215)
  - [x] `stacks/geospatial-.*json` (#212)
  - [x] `stacks/ml-cuda10.1-.*json` (#212)
- [x] Create commands that executes `buildx bake` commands continuously while changing targets. Update the `Makefile`. (#220)
- [x] Switch `docker-compose build` to `docker buildx bake` in GitHub Actions build workflow. (#220, #226)
- [x] More [OCI pre-defined annotations](https://github.com/opencontainers/image-spec/blob/main/annotations.md). (I think it would be useful to be able to show these in the reports.)
  - [x] `org.opencontainers.image.revision` (#220)
  - [x] `org.opencontainers.image.version` (#227)
  - [x] `org.opencontainers.image.title` (#227)
  - [x] `org.opencontainers.image.description` (#227)
  - [x] `org.opencontainers.image.base.name` (#227)
- [x] Multi-architecture (amd64 & arm64) build. (#242)
- [x] Add tests for `docker-bake.json` files. (Test if the bake files generated from the stack files are correct with `bake --print` command) (#239)

eitsupi:
FYI, I mentioned in https://github.com/rocker-org/rocker-versioned2/pull/214#issuecomment-896990256 that it is difficult to merge the currently separated stack files into `X.Y.Z.json` for build convenience, but after we use the `bake` command to build, the GitHub Actions restriction will be removed.

For example, if we want to extract only `r-ver`, `rstudio`, `tidyverse`, and `verse` from `devel.docker-bake.json` for daily build, we only have to prepare a file `build/develdaily.hcl` like the following:

```hcl
group "default" {
    targets = ["r-ver", "rstudio", "tidyverse", "verse"]
}
```

By merging this file with `devel.docker-bake.json`, only the default targets will be extracted.

```shell
$ docker buildx bake --print -f tmp/develbuild.hcl -f bakefiles/devel.docker-bake.json
[+] Building 0.0s (0/0)                                                                                                               
{
   "target": {
      "r-ver": {
         "context": "./",
         "dockerfile": "dockerfiles/Dockerfile_r-ver_devel",
         "tags": [
[...]

Issue #181, opened 2021-06-20
[Tracking] Build system automation


eitsupi:
As the issue #162 has become longer and there are more topics related to automation other than the automatic Dockerfile update that was the original topic, I would like to track the updates related to automation in this new issue.

- [x] Automated update of definition files.
  - [x] Automated update Dockerfile when new version of R / RStudio is released.
    - [x] `rocker/r-ver`, `rocker/rstudio`, `rocker/tidyverse`, `rocker/verse` (#164)
    - [x] `rocker/shiny`, `rocker/shiny-verse` (#175)
    - [x] `rocker/geospatial`, `rocker/binder` (#175)
    - [x] `rocker/ml`, `rocker/ml-verse` (#192)
    - [x] others (#227)
  - [x] GitHub Actions ([Dependabot for GitHub Actions](https://dependabot.com/github-actions/)) (#233)
- [x] Automated build and push container to DockerHub.
  - [x] Automated builds of `latest` tagged images when dockerfiles or scripts are changed.
    - [x] `rocker/r-ver`, `rocker/rstudio`, `rocker/tidyverse`, `rocker/verse` (#169)
    - [x] `rocker/shiny`, `rocker/shiny-verse` (#175)
    - [x] `rocker/geospatial`, `rocker/binder` (#175)
    - [x] `rocker/ml`, `rocker/ml-verse` (#192)
  - [x] Automated builds of non-latest images.
    - [x] `rocker/r-ver`, `rocker/rstudio`, `rocker/tidyverse`, `rocker/verse`, `rocker/geospatial`, `rocker/shiny`, `rocker/shiny-verse`, `rocker/binder`, `rocker/cuda`, `rocker/ml`, `rocker/ml-verse` (#223)
    - [x] others (#227)
  - [x] Automated builds of higher level taged (e.g. `4`, `4.0`, `4.1`) images
    - [x] Set shared tags in stack files. (#214)
    - [x] Build images with these tags. (#220, #226)
  - [x] Scheduled daily builds of devel images.
    - [x] R-devel: `rocker/r-ver:devel`, `rocker/rstudio:devel`, `rocker/tidyverse:devel`, `rocker/verse:devel` (#180)
    - [x] RStudio daily build: `rocker/rstudio:latest-daily`, `rocker/tidyverse:latest-daily`, `rocker/verse:latest-daily` (#194)
- [x] Changing the structure of definition files (tasks to simplify the workflow to be automated).
[...]

PR #206, opened 2021-08-04
Add reports generation system

eitsupi:
Suggested by https://github.com/rocker-org/rocker-versioned2/issues/181#issuecomment-891075725

Add a script and a report template to create reports about the contents of the container images.
By publishing the generated Markdown files on GitHub (maybe on the wiki?), users will be able to go back and examine past images.

The report will look like this:

image

To create reports, we will need files in `tmp` directory about the container images created by the following command. (need docker)

```shell
$ make inspect-image-all
```

Next, run the following command to generate reports in the `reports` directory. (need R 4.1 and the tidyverse packages etc., I run this command in `rocker/tidyverse:latest`)

```shell
$ make report-all
```

**ToDo**

- [x] Add jobs to the Makefile
- [x] Python packages list
- ~~commit hash~~ Cannot be implemented at this time because it is necessary to set [the `org.opencontainers.image.revision` annotation](https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys) when building images. I would like to set this up after we change the build method to one using `buildx bake`. (#199)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@
 .Ruserdata
 *.Rproj
 .*.swp
+
+tmp
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,6 @@
 .*.swp
 
 tmp
+
+# Use for build reports
+/reports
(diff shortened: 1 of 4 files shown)


eitsupi:
@cboettig @noamross Now we can generate simple reports, what do you think about this?

If you would like to adopt the idea of publishing these reports on the wiki, I think we can do so by incorporating the following functionality into the GitHubActions' workflows.

1. After the `docker build` is finished, run `make inspect-image-all` to store the files in the `tmp` directory.
2. Use the `upload-artifact` action to store the `tmp` directory.
3. Start a new job in the `rocker/tidyverse:latest` container.
4. Check out this repository.
5. Use the `download-artifact` action to restore the `tmp` directory.
6. Check out the wiki in the `reports` directory.
7. Run `make report-all`.
8. Use the [`stefanzweifel/git-auto-commit-action`](https://github.com/stefanzweifel/git-auto-commit-action) to add new reports to the wiki.

cboettig:
@eitsupi this looks great.  Yes, I think adding the GitHubActions workflow to publish to the wiki would be a great idea.  
I'll merge this as is and then we can take a look at the Wiki integration in as a separate PR.

PR #211, opened 2021-08-07
Publish build reports to the wiki

eitsupi:
This PR is a continuation of #206 and was motivated by https://github.com/rocker-org/rocker-versioned2/issues/181#issuecomment-891075725.

Automatically publish reports on the Wiki after the container image build and update the Wiki's Home.
We will be able to find out what packages are included in every image that is built, and it will be easier to reuse old images.

Please activate the Wiki when merging this PR.

## How file generation works?

image

The individual reports are as described in https://github.com/rocker-org/rocker-versioned2/pull/206#issuecomment-894333827.

The HOME of the Wiki is generated by the newly added RMarkdown file and Makefile jobs in this PR. (`make wiki-home`)
The wiki repository stores tsv files with names specific to each build workflow and each R version, and when `Home.md` is generated, all tsv files are read and the image list is updated.
We can trace the past images by checking the commit history in `Home.md`.

image

### Wiki Home

image

### Reports

image

## ToDo

- [x] Test for GitHub Actions on fork
  - [x] Publish reports to the wiki after image build.
  - [x] Update the wiki's Home.
- [x] New job to update `Home.md`

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -35,7 +35,7 @@ jobs:
     needs: generate_matrix
     runs-on: ubuntu-latest
     strategy:
-      fail-fast: true
+      fail-fast: false
       matrix: ${{fromJson(needs.generate_matrix.outputs.matrix)}}
     steps:
       - uses: actions/checkout@v2
@@ -82,7 +82,7 @@ jobs:
           docker image push rocker/ml-verse:latest
       - name: Inspect built image
         run: |
-          make inspect-image-all
+          IMAGELIST_NAME=${{ matrix.r_version }}.tsv make inspect-image-all
       - name: Upload artifacts
         uses: actions/upload-artifact@v2
         with:
@@ -90,8 +90,11 @@ jobs:
           path: tmp
 
   publish_reports:
-    needs: build
+    needs: [generate_matrix, build]
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix: ${{fromJson(needs.generate_matrix.outputs.matrix)}}
     container:
       image: rocker/tidyverse:latest
     steps:
@@ -107,9 +110,10 @@ jobs:
         with:
           name: tmp
           path: tmp
-      - name: Generate reports
+      - name: Generate reports and update wiki home
         run: |
           make report-all
+          IMAGELIST_NAME=${{ matrix.r_version }}.tsv make wiki-home
       - name: Update wiki
         uses: stefanzweifel/git-auto-commit-action@v4
         with:
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -112,4 +112,4 @@ jobs:
         uses: stefanzweifel/git-auto-commit-action@v4
         with:
           commit_message: Automated update
-          repository: wiki
+          repository: reports
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -86,6 +86,7 @@ jobs:
       - name: Upload artifacts
         uses: actions/upload-artifact@v2
         with:
+          name: tmp
           path: tmp
 
   publish_reports:
@@ -104,6 +105,7 @@ jobs:
       - name: Download artifacts
         uses: actions/download-artifact@v2
         with:
+          name: tmp
           path: tmp
       - name: Generate reports
         run: |
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -90,11 +90,8 @@ jobs:
           path: tmp
 
   publish_reports:
-    needs: [generate_matrix, build]
+    needs: build
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-      matrix: ${{fromJson(needs.generate_matrix.outputs.matrix)}}
     container:
       image: rocker/tidyverse:latest
     steps:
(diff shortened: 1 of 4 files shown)


eitsupi:
Sorry, I meant to create a PR for my fork, but accidentally created it here...
I'll test it on my fork and come back.

eitsupi:
In this PR, the report creation is done in the same workflow (`core.yml`) as the image build, but I think we can also create a separate workflow that pull all published images from the registry using the tag list and creates a report, and have it run periodically.
(In the future, `docker pull` will always be required before inspect, since the built image will not be left locally when multi-architecture builds are performed.)

Should I work on defining such a workflow before implementing regular updates of R4.0.0 - R4.0.5 version images?

By the way, R4.1.1 is scheduled to be released tomorrow, but the current GitHub Actions will not update the R4.1.0 images if the latest version is switch to R4.1.1, so the frozen image of `rocker/r-ver:4.1.0` will not be released forever.
When I started working on the upgrade automation, my intention was to start regular updates of the frozen images earlier to avoid this problem, but I was not able to do so in time for the release of R4.1.1......

cboettig:
Thanks, this looks great!

Yes, I would lean towards having this implemented as a separate workflow first?  (I always get nervous having multiple independent tasks in a workflow since it can make debugging and re-running a nuisance, even though the matrix structure tends to push us that way...).  


Okay, good point about the 4.1.1 release.  So we just need to tickle the 4.1.0 build so it switches to the locked version, or we need to manually update the 4.1.0 stack file first?

eitsupi:
@cboettig The stack files listed in #181 should be automatically updated and a PR will be created, so no manual updates are needed.
However, building and publishing not-latest images is not automatic with the current GitHub Actions configuration, so if you want to publish a frozen R4.1.0 image before implementing a build of a past version, you will need to build images and push them from your local machine (as you have been doing before).

I would appreciate it if you could merge this PR immediately, since the reporting jobs in this PR and the reporting jobs in another workflow that will be created in the future will not conflict. That way, the first report will be published tomorrow when the R version upgrade PR is merged and the workflow works.
Please check my fork for workflow behavior. https://github.com/eitsupi/rocker-versioned2/actions/runs/1113453206
I have reduced the number of images to build to shorten the build time, but other than that, it is the same as this PR

If you can wait a few days for R4.1.0 to be frozen, I can add a new workflow for inspecting published images and the ability to build all versions in parallel to `core.yml`.
So, if you are not in a hurry to freeze R4.1.0, there is no need to do anything other than merging the PR.

Thanks.

cboettig:
Nice, sounds like a plan. thanks!

eitsupi:
@cboettig Thank you for merging!

Can you please create the first page on the wiki for this repository? It seems that if the first page is not created, the wiki repository does not exist and cannot be updated.

eitsupi:
That works🎉
image

eddelbuettel:
That's _really_ slick.  I often pondered about 'cron'-alike jobs via GH Actions and the problem of where to store results.  Writing to the Wiki is pretty neat.

eitsupi:
Yes, we can only see Markdown files from the GitHub website, but if we operate from git, we can store any files (in fact, this workflow store tsv files), so it can be used for various things...

eddelbuettel:
> we can only see Markdown files from the GitHub website

Not exactly. I work a lot with GitHub Pages and it can of course serve whatever you want.  But altering the files is always a commit, and I have crojobs all the way down to the hourly level where that would be absurd.

eitsupi:
I apologize for the confusing wording.
When I described the "website" in my earlier post, I was referring to opening the GitHub wiki from a browser.
GitHub wiki ignores directory hierarchies and non-Markdown files.

If we check out the wiki, we will see that it contains the following files, but we will not be able to recognize the existence of `imagelist/4.1.1.tsv` from our browser.

```shell
$ tree .
.
├── 08f20b78c168.md
├── 48e38c2cca33.md
├── 851bb9c6ab25.md
├── 8cfc59aa5964.md
├── 8d553e1d71f5.md
├── 8f99394fb865.md
├── Home.md
├── bb3eb00fb52e.md
├── ceac7f61f320.md
├── d7e92973b2f3.md
├── effb07af3037.md
├── fbe72eb41afe.md
└── imagelist
    └── 4.1.1.tsv

1 directory, 13 files
```

image

Also, here all the md files are in the root of the repository, but even if there are md files in subdirectories, they all seem to be in the same hierarchy on the web.
The `jupyter/docker-stacks` wiki has such a structure.
https://github.com/jupyter/docker-stacks/wiki

eddelbuettel:
Thanks for the clarification. I also noticed these files as entries in the wiki 'table of contents' (which, as a wiki, is filename-based).  Can we make them not be sha1-based?  Or maybe prefix by `yyyymmdd-` ?

eitsupi:
> I also noticed these files as entries in the wiki 'table of contents' (which, as a wiki, is filename-based). Can we make them not be sha1-based? Or maybe prefix by `yyyymmdd-` ?

This is due to the use of the abbreviated image id output by `docker image ls -q` as the unique name for each image (and report file name).

https://github.com/rocker-org/rocker-versioned2/blob/7a06d1c2136f53b13b5ee3976751b179ac916316/Makefile#L78

I think it needs to be a unique name that does not include dates, etc., since the same image may be observed at different times.
It doesn't look great, but I feel it's searchable, as it's a quick hit when we search for an ID on the wiki. (I'm assuming we can look up the ID in the wiki Home history and search for it on the wiki.)

It may be possible to set up a sidebar (`_Sidebar.md`) and display it as an alias on the sidebar.

PR #212, opened 2021-08-09
Some fixes to use `make-bakejson.R`

eitsupi:
`make-bakejson.R` doesn't support some of the existing stack files, so fix it.
Specifically, align the IMAGE and TAG hierarchies in all stack files.

1. `stacks/geospatial-.*json`: Only the configuration change of the stack files, and it does not affect the Dockerfiles. `stacks/geospatial-ubuntugis.json` is split by R version, like other stack files. (As an aside, I think this will allow we to build geospatial-ubuntugis again with GitHub Actions.)
2. `stacks/ml-cuda10.1-.*json`: Even though `rocker/ml:XYZ` has exactly the same contents as `rocker/ml:XYZ-cuda10.1` and `rocker/ml-verse:XYZ` has exactly the same contents as `rocker/ml-verse:XYZ-cuda10.1`, multiple tags cannot be set in the build using docker-compose, so until now, we had to prepare different Dockerfiles and build different images. By tagging these images on GitHub Actions, `Dockerfile_ml_X.Y.Z` and `Dockerfile_ml-verse_X.Y.Z` are no longer needed and can be removed. When using docker-bake.json, we can set these tags at the same time in one build by setting the tags key.

~Also, modify `make-dockerfiles.R` and `write-compose.R` to ignore the tags and platforms keys that can be set for `docker-bake.json`.~ I'm sorry, we didn't need this!


cboettig:
LGTM :tada:

PR #216, opened 2021-08-12
Add new workflow to generate reports on all tagged images

eitsupi:
Add another workflow definition file `reports.yml` (and related files) that separate only the reporting part from `core.yml`, which builds container images.

Since we need to `docker pull` for all tags in this workflow, I added `docker-bake.json` files for tag extraction.
`docker-bake.json` will be used for build in the future, so we will be able to refer to the exact same tag in build and pull.
And, I have made minor modifications to `make-stacks.R` to generate a new matrix file ` build/matrix/all.json` that is required to run jobs in parallel for each R version. This can also be used in `core.yml`.

Currently, reports are generated on every build, so I believe that this workflow is rarely needed except for the first time when we need to create a built report prior to the automatic report creation.
Therefore, triggering is currently manual only.
Please run it only once after merging this PR.

Run results on my fork

image

image

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -5,7 +5,7 @@ on:
     branches:
       - "master"
     paths:
-      - "build/matrix/**"
+      - "build/matrix/latest.json"
       - "scripts/**"
       - "dockerfiles/Dockerfile_r-ver_*"
       - "dockerfiles/Dockerfile_rstudio_*"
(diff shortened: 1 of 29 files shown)


eitsupi:
On closer inspection, `rocker/r-ver:4.0.1` etc. were not in the list. I will investigate.

eitsupi:
Run `docker image prune` before the job solved the above problem.

image

eitsupi:
Running this workflow will keep records of past images, so next, I'd like to prepare for all versions of the build and use `docker buildx bake` to record the commit hash in the image annotation in future builds. (That will be recorded in `org.opencontainers.image.revision`, which is currently NA in reports)
Once we know the version of the image source, we can investigate the differences between the images in more detail.

cboettig:
Thanks @eitsupi , merged and triggered once manually now...

eitsupi:
🎉
image

PR #221, opened 2021-08-14
Change the Dockerfiles naming convention and remove unused files

eitsupi:
Fix #218, Fix #219

PR #223, opened 2021-08-15
Create releases and build non-latest images automatically

eitsupi:
Suggested by https://github.com/rocker-org/rocker-versioned2/issues/181#issuecomment-891075725

The workflow definition `core.yml` that currently builds the images will only build the latest version of the images.
Therefore, the previous version will not be frozen by GitHubActions forever.

GitHubActions allows we to change the behavior of each type of trigger.
Using this feature, automatically create a release when a new version of R is released, and trigger it to build the non-latest image with `core.yml`.

Release
image

The new workflow triggers non-latest image building workflow.
image

Once this PR is merged, in the future, merging the automatically created PRs associated with the release of a new version of R, such as #213, will also automatically build the old version of the image.
The maintainers does not need to do anything other than merge the bot's PRs.
Unfortunately R4.1.1 was released a few days ago, so you will need to run it manually the first time.
(Tags with the same name will not be overwritten, so the workflow will fail if triggered incorrectly.)

**ToDo**
- [x] Waiting for #220 and #221 merges
- [x] Waiting for the build workflow fixed (https://github.com/rocker-org/rocker-versioned2/pull/221#issuecomment-899902688, #226)
- [x] ~~[Set a personal access token](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token)~~ Solved by triggering `workflow_run` instead of `release`


eitsupi:
@cboettig @noamross Sorry, I forgot that a workflow using `GITHUB_TOKEN` cannot be used to trigger other workflows.
https://docs.github.com/en/actions/reference/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token

If you agree to implement a mechanism to build non-latest images in this release, could you please set a personal access token to this workflow?

eitsupi:
#226 will cause `core.yml` to trigger `reports.yml`, so I excluded Dockerfiles from `core.yml` triggers to prevent the PR merge from directly triggering `core.yml` when upgrading R (which includes new Dockerfiles and new `latest.json`).
This is because if the PR merge triggers `core.yml`, then the PR merge and the creation of the release will trigger `core.yml` twice, and then `reports.yml` will also be triggered twice, which may cause conflicts with wiki editing.

Since `stacks/devel.json` is also changed when Dockerfile is changed except when R is upgraded, the behavior of `core.yml` being triggered after the Dockerfile modification process in normal times remains the same.

eitsupi:
> 
> 
> #226 will cause `core.yml` to trigger `reports.yml`, so I excluded ~Dockerfiles~ versioned Dockerfiles from `core.yml` triggers to prevent the PR merge from directly triggering `core.yml` when upgrading R (which includes new Dockerfiles and new `latest.json`).
> This is because if the PR merge triggers `core.yml`, then the PR merge and the creation of the release will trigger `core.yml` twice, and then `reports.yml` will also be triggered twice, which may cause conflicts with wiki editing.
> 
> Since ~`stacks/devel.json`~ devel Dockerfile is also changed when versioned Dockerfile is changed except when R is upgraded, the behavior of `core.yml` being triggered after the Dockerfile modification process in normal times remains the same.

This change means that `rocker/rstudio:latest` will not be built even if you merge PRs when rstudio is upgraded, which occurs independently of R version upgrades, so the `rstudio_devel.Dockerfile` is also updated when RStudio is upgraded.

When a new version of RStudio is released, `rstudio_devel.Dockerfile` and `rstudio_X.Y.Z.Dockerfile` will be updated and `rocker/rstudio:latest` will be built.

eitsupi:
When you merge this PR, the latest version should be built due to the `rstudio_devel.Dockerfile` update, but the latest version will be built again when you create the release.
So I think it is better to manually cancel the build when you merge this PR, and then trigger the release workflow manually, then all images are built.

Sorry it's complicated!

eitsupi:
When I check the old stack files (with `make-stacks.R`), I find that the version of RStudio is not fixed in some files. 
Also, `CTAN_REPO` is not fixed in `ml-verse`.
I think it's better to fix these parts, how about it? 
If you like, I would like to fix it in this PR. 

https://github.com/rocker-org/rocker-versioned2/blob/a339ed1d8e928a613d4645279e6d4c0266fbc3dd/dockerfiles/ml_4.0.0.Dockerfile#L9
https://github.com/rocker-org/rocker-versioned2/blob/a339ed1d8e928a613d4645279e6d4c0266fbc3dd/dockerfiles/ml_4.0.1.Dockerfile#L9
https://github.com/rocker-org/rocker-versioned2/blob/a339ed1d8e928a613d4645279e6d4c0266fbc3dd/dockerfiles/ml_4.0.3.Dockerfile#L9
https://github.com/rocker-org/rocker-versioned2/blob/a339ed1d8e928a613d4645279e6d4c0266fbc3dd/dockerfiles/rstudio_4.0.3.Dockerfile#L9
https://github.com/rocker-org/rocker-versioned2/blob/a339ed1d8e928a613d4645279e6d4c0266fbc3dd/dockerfiles/ml_4.0.3.Dockerfile#L9
https://github.com/rocker-org/rocker-versioned2/blob/a339ed1d8e928a613d4645279e6d4c0266fbc3dd/dockerfiles/ml_4.0.4.Dockerfile#L9

cboettig:
Good catch, yes, makes sense to fix those here.

No worries, I should be able to merge, cancel it's build, and then trigger the release build.

eitsupi:
Completed fixing the `RSTUDIO_VERSION` and `CTAN_REPO` of the old stack files.
And, although it was not the original purpose of this PR, it was incidental, so I also modified `install_rstudio.sh` so that RStudio Daily Build can be installed.

When you merge this PR and manually run "Create new release", the three workflows are executed in sequence as below.
(The build workflow is failing because my fork can't push the Docker image to DockerHub)

image

image

image

image

eitsupi:
@cboettig Thank you for merging. Probably the workflow you run manually is different?
image

cboettig:
Ok, queued https://github.com/rocker-org/rocker-versioned2/actions/runs/1149208692

eitsupi:
@cboettig Thanks. It seems to have worked properly, so it would be better to cancel this one (and the workflow of the report that runs after this one is canceled).
image
It will work automatically next time, so please just this once.

eitsupi:
I was worried about conflicts when updating the wiki if it was built twice in close time, but it seems to have been updated successfully this time.

PR #226, opened 2021-08-17
Fix the buildx bake workflow

eitsupi:
Fix https://github.com/rocker-org/rocker-versioned2/pull/221#issuecomment-899902688
Modify the workflow for building container images using buildx configured in #220 to work.

With `buildx`, the built image is only saved either to the local Docker with the `--load` option or to the registry with the `--push` option.
The `docker-compose build` that we have been using so far stores the image locally, so I added the `--load` option in #220 without understanding that we can't use `--push` and `--load` at the same time.
In this PR, the `--load` option is removed.

With the `--load` option gone, the built image is no longer left in the Docker that built the image.
Therefore, I removed the reporting function from `core.yml`, which assumes that there are images built on the local Docker, and consolidated the reporting function into `reports.yml` so that `reports.yml` is executed after `core.yml` is executed.
(I was planning to make such a change in the future, since the `--load` option is not available for multi-arch images, but it turned out that `--push` and `--load` are not compatible, so I did it this time)


cboettig:
nice, this makes sense, thanks for explaining

PR #227, opened 2021-08-17
Merge and change the structure of stack files and update build strategy, and some improvements

### LINKED ISSUE #230: Change the copy order according to best practices?
Currently, the Dockerfiles of `rocker/r-ver` has the following structure, so when one of the files in `scripts` is updated, the subsequent layers will be new even if the cache is available.

```Dockerfile
FROM ubuntu:focal

COPY scripts /rocker_scripts
RUN /rocker_scripts/install_R.sh
```

According to [the best practices for writing Dockerfiles](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#use-multi-stage-builds), the following configuration is more desirable as it increases the likelihood of sharing layers with each build. If the base image `ubuntu:focal` has not been updated and `install_R.sh` has not been changed, the new layer will be only the last `COPY`, whose build time and the new layer size is very small.
(Of course, `rocker/r-ver:devel` expects the result of `install_R.sh` to be different every time, so we need to disable the cache.)

```Dockerfile
FROM ubuntu:focal

COPY scripts/install_R.sh /rocker_scripts/install_R.sh
RUN /rocker_scripts/install_R.sh
COPY scripts /rocker_scripts
```

This change and the enabling of caching (which can be controlled in `docker-bake.json`, e.g. `"cache-from": ["rocker/r-ver:X.Y.Z"]`) will shorten the build time and slightly reduce the pull time.
Especially for multi-architecture builds, the execution time of `install_R.sh` can be very long, so this change is very worthwhile from the perspective of reducing the build time.

### LINKED ISSUE #231: Set `DEFAULT_USER` as ENV on `rocker/rstudio`'s Dockerfiles?
The non-root user that exists in `rocker/rstudio` and its derived images is created in the following process, but the name `rstudio` is not visible in the Dockerfiles, so we need to read the shell scripts thoroughly or run and check the containers.

https://github.com/rocker-org/rocker-versioned2/blob/30fca60a9241fc1e3abc8ec9f16d4639f4343f29/scripts/default_user.sh#L7-L8

It will be easier for users to understand if `DEFAULT_USER` is displayed on the Dockerfiles.

### LINKED ISSUE #232: New RStudio version name (Preview and Daily build)
https://github.com/rstudio/rstudio/pull/9695

Following #202 and #228, it seems that the naming convention has changed again.
This time it also affects the preview version and is mentioned in the release notes, so it is expected that it will not change for a while.
https://www.rstudio.com/products/rstudio/download/preview-release-notes/

The current file names are as follows.

- rstudio-server-2021.11.0-daily+17-amd64.deb
- rstudio-server-2021.09.0-preview+338-amd64.deb

eitsupi:
Completely merge the stack files that exist for each version into `X.Y.Z.json`(version-fixed images) or `extra.json`(special images that build only the latest version; geospatial dev images).
To shorten the build time, change the structure of the matrix used in GitHub Actions so that images that do not have dependencies within the same version can be set as separate groups and built in parallel for each group.
To do such a build on GitHub Actions, modify the structure of the matrix files to include the `group` on `docker-bake.json`.
The new matrix files are automatically generated from `docker-bake.json` files.

image

image

Some image names (`IMAGE` in stack files, which is used as Dockerfiles' name etc.) changed in order to avoid duplicate key names in `docker-bake.json`, the compose files will not be generated normally from the integrated stack file, so update the `write-compose.R` not to generate compose files from the integrated stack files.
This also affects some Dockerfiles' name.

And, this PR makes some small improvements to stack files or files generated from stack files.

- Add some annotations to container images.
- Change to use cache when building `rocker/r-ver:X.Y.Z` etc. (close #230)
- Set `DEFAULT_USER` to ENV for  `rocker/rstudio` etc. (close #231)
- Update `install_rstudio.sh` for RStudio new calendar based versioning. (close #232)

**ToDo**

- [x] Set `group` in each stack files and `docker-bake.json` files.
- [x] Add a script generate new matrix files, which contains `group`.
- [x] Merge stack files to `X.Y.Z.json` or `extra.json` (geospatial dev images) and to be updated by ~~`sync-template-vars.R` and~~ `make-stacks.R`.  
  _Note_: Due to the trigger of GitHub Actions, `make-stacks.R` now also updates `devel.json` which is a template of `X.Y.Z.json` (#223), so there is no point in separating `sync-template-vars.R` and `make-stacks.R`, I integrate `sync-template-vars.R` functionality into `make-stacks.R`.
  - [x] cuda11
  - [x] ubuntu18.04
  - [x] geospatial ubuntugis
  - [x] geospatial dev-osgeo
- [x] Add a new workflow `extra.yml` to build images defined by `extra.json`.
- [x] Add anotations in `docker-bake.json`
  - [x] `org.opencontainers.image.version`
  - [x] `org.opencontainers.image.title`
  - [x] `org.opencontainers.image.description`
  - [x] `org.opencontainers.image.base.name`
- [x] Divide the `COPY` section and the `RUN` section in the stack files and Dockerfiles into two parts, and use cache when building images.
- [x] Set `DEFAULT_USER` to ENV.
- [x] Add new builds job which build each group to the `Makefile`, and use them in GitHub Actions workflow.
- [x] Workflow tests on my fork.


eitsupi:
As for the build time of `ubuntugis`, if I remember correctly when I looked at the log before it disappeared, it was about 30 minutes, so I think it won't take more than 3 hours if I add the current build time of less than 2 hours.
Multi-architecture builds of `rocker/r-ver` will take an hour and a half, but I don't think it will take more than 5 hours total, so we should be able to build it.
(GitHubActions are capped at 6 hours.)

By the way, when the arm64 version of RStudio Server is released, it probably won't even be able to muli-arch build up to `rocker/verse` unless RSPM provides the arm64 binary...

eitsupi:
In 32b1548, I have added `labels` to the devel stack file, which is a description of the container images.
This will be transcribed into `docker-bake.json` and written to the image in the same way as `LABEL` in Dockerfiles.
(I think it would be useful to be able to show them in reports in the future.)

I used the explanations on Docker Hub, but I would appreciate it if you could give me instructions if there is a more appropriate text.

eitsupi:
This PR has become very huge as the stack files and almost all the files generated from them have changed.
However, most files (especially Dockerfiles) have only minor changes, so I'd appreciate it if you could check and merge them.

This should be the last huge PR in a series of build system automation (#181), and I think that automation is almost over when it is merged.

eitsupi:
I didn't understand what the `geospatial-dev-osgeo` was for, so I thought it was done in its current state, but I noticed that it was mentioned in https://github.com/rocker-org/rocker-versioned2/pull/214#issuecomment-896965328.
`geospatial-dev-osgeo` doesn't currently have a version and only the latest image exists. Should I follow this and include it only in the latest stack file (currently `4.1.1.json`)?
Or should I embed it in `devel.json` and transfer a copy to the latest two versions, just like any other?

I also left `core-4.0.0-ubuntu18.04.json` and `geospatial-4.0.0-ubuntu18.04.json`, but these can be incorporated into `4.0.0.json` as a new group. Is it a good idea to do this?

eitsupi:
It's a late question, is there any point in continuing to build old version of `geospatial-ubuntugis`?

Unlike other images, the default CRAN is not date-fixed, so I think it is better to separate `geospatial-dev-osgeo` and `geospatial-ubuntugis` from the stack `X.Y.Z.json` for each version and build only the latest version and build them at the base timing (eg once a month), like `rocker/rstudio:latest-daily`.

eitsupi:
To recap, I would prefer to make the following modifications.

- Integrate `core-4.0.0-ubuntu18.04.json` into `4.0.0.json`. Thus, it will be built for each release of R.
- `geospatial-dev-osgeo.json`, and `geospatial-ubuntugis` in `X.Y.Z.json` together to make `extra.json`. `extra.json` is updated by `make-stacks.R` to keep up with the latest R version.
- Create a new workflow `extra.yml` to build images from `extra.json` and do scheduled builds on the first day of each month. The base image is `rocker/verse:X.Y.Z`.

Originally, I should have thought this far before starting the work... Do you think these changes make sense?

cboettig:
Nice, these proposed changes make sense to me.  yeah, sometimes the better way to do things is only obvious after trying some alternatives!

cboettig:
Oh, and in reply to the above -- right, I don't think it makes sense to build old versions of the geospatial dev images (ubuntugis and dev-osgeo).  Maintaining those may be tricky if those sources change (though I think our config is relatively stable for them), and dev-osgeo in particular is an intensive build.  We might drop ubuntugis if dev-osgeo is proving stable, since these two are a bit redundant.  Still, many users and CI platforms elsewhere use the PPA-based ubuntugis strategy it's faster (though not being able to use RSPM binaries somewhat defeats that anyway), and so it can be nice to be able to replicate that.  On the flip side the value-add of building the latest osgeo sources is much larger, since they are both slow and somewhat tricky to configure properly to talk to the R packages...

eitsupi:
I changed the structure of the stack file as suggested above and also set up automatic updates and build workflow.

image

image

cboettig:
@eitsupi This looks good to me.  I think we're ready to merge?

eitsupi:
@cboettig Thank you for your review.
Yes, I think it will work fine if you merge it.

Note that merging this PR will trigger the release workflow, but ignore it as the workflow will be canceled as the tag `R4.1.1` already exists.

PR #236, opened 2021-09-09
Update wiki home to show all images

eitsupi:
Update the wiki template to match the build changes made in #227

Change the list of images displayed in the wiki home to be generated from docker-bake.json, so that all images are displayed in the table.
Also, make minor changes to the appearance. (to make it easier to see the history differences, plane text)

Change appearance(Remove hyperlinks at `ImageName`, `ID`s are no longer codes)

image

Show cuda11 images, 4.0.0-ubuntu18.04 images, extra images.

image

image

### DIFF
--- a/.github/workflows/reports.yml
+++ b/.github/workflows/reports.yml
@@ -1,9 +1,15 @@
 name: Report published images
 
 on:
+  push:
+    branches:
+      - "master"
+    paths:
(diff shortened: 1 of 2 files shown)


eitsupi:
Oops, the R versions of the extra images are strange. Modify so that the R versions are not displayed.

PR #238, opened 2021-09-10
Change the naming convention for reports uploaded to the wiki

eitsupi:
The filenames of the reports currently saved as wiki pages are the image IDs.
These filenames are hard to guess when displayed in search results on GitHub.

image

Therefore, by this PR, the file names will be changed to include the image names. The new report will appear in the search results as follows:

image

Renaming the files of the reports is done in the process of generating the wiki home, but the appearance of the wiki home is not changed in this PR.


cboettig:
nice, sounds good to me.

eitsupi:
Thank you for merging. It worked fine🎉

image

PR #241, opened 2021-09-12
Update some rocker scripts for arm64 build (and some minor improvements)

eitsupi:
Related to #144, continuation of work of #195.

Modify the following script to work with arm64 builds.

- `install_s6init.sh`
- `install_texlive.sh`
- `install_pandoc.sh`

With this improvement, all software required to build `rocker/verse` except RStudio Server can be installed on arm64 images.

According to my research, `install_wgrib2.sh` cannot be executed with arm64, so `rocker/geospatial` will not be able to make an arm64 build even if the arm64 version of RStudio Server appears.
The survey results were added as comments in the script.

In addition, the following minor improvements have been made in this PR.

- Remove S6 installer in `install_s6init.sh`.
- Avoiding the error of running `install_texlive.sh` without the non-root user. (In most cases, it's a script that runs after installing RStudio Server and create the default user, so it hasn't been a problem until now.)
- Addition of BiocManager installation that was missing in `install_geospatial.sh`. (In most cases, it's a script that runs after installing BiocManager, so it hasn't been a problem until now.)
- Set `-ncpus -1` argument for all `install2.r` commands.


cboettig:
This looks wonderful, nice work!  :tada:

(yeah, no worries about wgrib2, that's a bit niche in geospatial as you probably gathered anyhow)

PR #242, opened 2021-09-12
Add rocker/r-ver:4.1.X arm64 build (multi architecture images)

eitsupi:
Related to #144
Add linux/arm64 to the `rocker/r-ver` platform and do multi-arch builds on GitHub Actions.

Currently, images built from `rocker/r-ver` (`rocker/rstudio` and `rocker/shiny`) cannot be arm64 built because there is no arm64 build of RStudio Server or Shiny Server.
So at this stage only `rocker/r-ver` can add arm64 builds.
And, we cannot install R binary packages from RStudio Package Manager with arm64 builds now. (#195)
Please consider the arm64 build as experimental at this time.

When I switched the build system to `docker buildx bake` in #220, I assumed that we could add the platform via `docker-bake.override.json`. However, considering the situation where arm64 builds will be enabled from a certain RStudio version in the future, I thought it would be more appropriate to set it up on individual stack files and set it up that way in this PR.


cboettig:
yup, this seems like a good way to handle the multi-arch.

PR #246, opened 2021-09-24
Report template update for multi-architecture images

eitsupi:
The current reports do not contain information about the architecture of the target image, so I added this.
The multi-architecture images report also adds a list of available architectures.

Currently, there is no multi-architecture built image from this repository yet, so here is the result of testing with `r-base`.

image

The report on images for amd64 only builds does not show a list of platforms.

image

**ToDo**

- [x] Test GitHub Actions workflow on my fork.


cboettig:
nice! LGTM

PR #250, opened 2021-09-28
Update for new RStudio versioning, download URL, and bump RStudio Server to 2021.09.0+351 in R 4.1.1 images

eitsupi:
A new version of RStudio `2021.09.0+351` was released today, but due to changes in the tag naming convention and download URL, it could not be installed with the current installation script, and the automatic update did not detect the new version.

The following changes are required

- Replace `+` in version name with `%2B` for download URL.
- Update the base URL.
- The rstudio/rstudio GitHub repository tags naming changed from "vX.Y.Z" to "YYYY.MM.X+Z", so GitHub APIv3 couldn't get the latest tag because of sorting, so we need to get the tags with git (in R, the `gert` package).

**ToDo**

- [x] Update the definition files.
- [x] Install test.
  - [x] RStudio 1.4.1717
  - [x] RStudio 2021.09.0+351
  - [x] daily
  - [x] preview

### DIFF
--- a/build/make-stacks.R
+++ b/build/make-stacks.R
@@ -14,6 +14,7 @@ library(purrr, warn.conflicts = FALSE)
 library(glue, warn.conflicts = FALSE)
 library(tidyr)
 library(stringr)
+library(gert)
 
 
 .latest_rspm_cran_url_linux <- function(date, distro_version_name, r_version) {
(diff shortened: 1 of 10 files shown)


eitsupi:
Ah, when I checked it now, the download URL was changed only for the release version, and daily and preview are the same as the original URL. It needs to be fixed...

cboettig:
hehe changing conventions always fun. Looks good, thanks for this!

PR #253, opened 2021-10-03
Add `install_quarto.sh`

eitsupi:
Add a script to install Quarto, which integrated with RStudio in the latest version of RStudio, 2021.09.

Do you think it's better to include this script in `rocker/rstudio`, `rocker/tidyverse`, or `rocker/verse`?

**Note**: The {quarto} R package was published in 2021-02-09, so execution of this script will fail in images with old CRAN mirrors.


cboettig:
Thanks!  I think I'd add it to `verse`, which is largely `tidyverse` + {publishing tools} already, so that fits well with the theme.

eitsupi:
I Added `install_quarto.sh` to the latest two `rocker/verse`.
The environment variable `QUARTO_VERSION` is currently set to `latest` to install the latest version of quarto-cli, but if a destructive change is made that causes a compatibility problem with the frozen version `{quarto}` R pakage (version 1.0), need to specify the version.

cboettig:
Nice!  We should probably have a NEWS.md kind of file to document changes like this that would be interesting to downstream users.  Probably should be distinct from the changelog that is more developer facing (i.e. the pull request tab is a pretty good change-log, but largely more developer oriented)

eitsupi:
I've added NEWS.md. Is it as you wish?

PR #256, opened 2021-10-06
Export build cache of `rocker/r-ver` to reduce build time

eitsupi:
I was wondering that builds after #227, which I intended to have the cache enabled, wouldn't reduce the build time at all (https://github.com/rocker-org/rocker-versioned2/issues/230#issuecomment-918334027), but I finally realized today that we need to explicitly export the cache...
https://github.com/moby/buildkit#export-cache

With this change, the layer that installs R this time will be cached on Docker Hub, and `rocker/r-ver` will complete the build instantly until the base image is updated.
By reusing the "install R" layer, users may also spend less time updating images with `docker pull` command.

### DIFF
--- a/bakefiles/4.0.0.docker-bake.json
+++ b/bakefiles/4.0.0.docker-bake.json
@@ -49,6 +49,9 @@
       ],
       "cache-from": [
         "docker.io/rocker/r-ver:4.0.0"
+      ],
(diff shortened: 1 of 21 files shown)


cboettig:
thanks for the ping, got buried in my list.  This looks good to me.

PR #263, opened 2021-10-13
first pass of gpu testing added to config_R_cuda.sh

RobJY:
This commit adds a subset of tests from the stand alone gpu test script that tests for gpu driver, toolkit, nvblas and tensorflow from within config_R_cuda.sh.

Is this:
1) the level of testing you were looking for in build tests?
2) the correct place to put build tests?

The output goes to /var/log/nvblas.log and still needs to be cleaned up.  I just wanted to start the conversation about how I should incorporate tests for the build.

### DIFF
--- a/scripts/config_R_cuda.sh
+++ b/scripts/config_R_cuda.sh
@@ -32,55 +32,7 @@ NVBLAS_GPU_LIST ALL' > /etc/nvblas.conf
 
 echo "NVBLAS_CONFIG_FILE=$NVBLAS_CONFIG_FILE" >> ${R_HOME}/etc/Renviron.site
 
-## Test GPU fuctionality
-TEST_FAIL=false
-# driver
-PROC_DRIVER_FILE=/proc/driver/nvidia/version
-if [ ! -f "$PROC_DRIVER_FILE" ]
-then
-  echo "$PROC_DRIVER_FILE doesn't exist" | tee -a /var/log/nvblas.log
-  echo "WARNING: CUDA driver may not be correctly installed." | tee -a /var/log/nvblas.log
-  TEST_FAIL=true
-fi
-# toolkit
-if ! TOOLKIT_CHECK_OUTPUT=$(nvcc -V 2>&1);
-then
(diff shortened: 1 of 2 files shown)


scripts/config_R_cuda.sh
```diff
@@ -32,6 +32,8 @@ NVBLAS_GPU_LIST ALL' > /etc/nvblas.conf
 
 echo "NVBLAS_CONFIG_FILE=$NVBLAS_CONFIG_FILE" >> ${R_HOME}/etc/Renviron.site
 
+test_config_cuda.sh /var/log/nvblas.log
```
eitsupi, line 35: It seems that this line caused the latest build to fail.
Isn't it `/rocker_scripts/test_config_cuda.sh /var/log/nvblas.log` correctly?

scripts/test-config-cuda.sh
```diff
@@ -0,0 +1,53 @@
+$LOG_LOC#!/bin/bash
```
eitsupi, line 1: Is this correct for the shebangs?

tests/gpu/misc/hosts
```diff
@@ -1,2 +1,2 @@
 [rocker-server-aws]
-54.81.189.38 ansible_ssh_private_key_file=~/Projects/Rocker/Rocker-aws-1/rocker-gpu.pem ansible_ssh_user=ubuntu
```
eitsupi, line 2: What does this IP address represent? It seems to change frequently. Does it make sense to specify it in a way that is not managed by git? (Including this file in gitignore and managing the template with git instead)

RobJY: Good point.  It's just for my Ansible build, so I'll add it to `.gitignore`

scripts/config_R_cuda.sh
RobJY: Oh, right, I was thinking it didn't need the parent directory because they're at the same level.  Sorry about that.
Looks like the directory is called `scripts`.
I've changed as you suggest to:
`/scripts/test_config_cuda.sh /var/log/nvblas.log`

PR #272, opened 2021-10-20
Cuda build fixes

RobJY:
I just created a markdown document to record details about attempted build fixes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 .Ruserdata
 *.Rproj
 .*.swp
+tests/gpu/misc/hosts
 
 tmp
 
--- a/tests/gpu/misc/build-fix-notes.md
+++ b/tests/gpu/misc/build-fix-notes.md
@@ -1,7 +1,5 @@
 # Build combination notes
 
-I'll just use this text file to store notes about build issues so I can create a PR and we can start discussing
-
 ## Build combination summary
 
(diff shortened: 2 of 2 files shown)


cboettig:
I think we should mark that combination as unsupported. Most users should have the option of using a newer tensorflow with cudnn8, right?  

I wonder if it would be worth adding the list of working configurations as well to the markdown document, perhaps with deep links to the test code which evidences that these configurations work.  Such documentation is always nice and an easier entrypoint than reading the tests directly.  

btw check out the lint check.

RobJY:
I've added a summary table to the notes.  The results are the same for the two driver versions I tested so the results collapse to a 2D table.  Please let me know if you'd like any changes.

I don't have a link to the successful builds since I build a Docker container based on a toolkit version and then cycle through the versions of tensorflow in my test script.  I could provide cleaned up Dockerfiles for the combinations that work with instructions for building.  How does that sound?

cboettig:
Thanks @RobJY ,  this looks good.  

On the tensorflow tests, why do we test 1.5 but not a more recent version in the 1.x series?  i.e. I think it might make sense to have the tests for the most recent version of 1.x (1.15 I believe), for software that still relies on the 1.x API before the compatibility-breaking changes of 2.x were introduced?

RobJY:
Sorry, that was a typo.  I only test the last update of version 1 which is 1.15 as you say.

Issue #267, opened 2021-10-16
Install the duckdb package with `install_tidyverse.sh` ?


eitsupi:
I think [the duckdb package](https://github.com/duckdb/duckdb/tree/master/tools/rpkg) is a worthwhile addition to `install_tidyverse.sh` because it can be used in combination with dbplyr and Apache Arrow.

https://github.com/rocker-org/rocker-versioned2/blob/7d8da9848c1cff0932dde04b6079229797815eaf/scripts/install_tidyverse.sh#L30-L40

However, the duckdb package was registered to CRAN on 2020-09-10, so cannot be installed from CRAN mirrors set in R<=4.0.1 images.
It seems that the command to install R packages (currently `install2.r`) needs to be changed as suggested in https://github.com/rocker-org/rocker-versioned2/issues/181#issuecomment-893659599.

PR #275, opened 2021-10-23
Include patched `install2.r` script for non-latest images

eitsupi:
Changes suggested in https://github.com/rocker-org/rocker-versioned2/issues/181#issuecomment-893659599 and https://github.com/rocker-org/rocker-versioned2/issues/267#issuecomment-948112263.

The `install2.r` command currently used in `rocker/r-ver` is from [the `littler` package](https://github.com/eddelbuettel/littler), so it is taken from CRAN snapshots and will not be updated in the non-latest images.

This PR adds `install2.r` scripts to rocker_scripts to allow older images to use the latest `install2.r` commands and to add new features.
First, add a new `--skipmissing` option to allow #267 to be resolved.

```shell
$ /rocker_scripts/bin/install2.r --error --skipmissing --skipinstalled -n -1 foo
```

In the future, I hope to be able to set up a GitHub Actions workflow to capture upstream changes automatically, but currently we have to edit them manually.

**ToDo**

- [x] Add a new script `patch_install_command.sh` to rewrite the `/usr/local/bin/install2.r` simlink on `rocker/r-ver`.
- [x] Merge the latest upstream `install2.r` (https://github.com/eddelbuettel/littler/pull/91).
- [x] Update NEWS.md


eitsupi:
Since `install_R.sh` can be run without dependencies, if we want to install R on an Ubuntu-based image that is not used in this repository, we can easily install R from source by copying and running `install_R.sh` only.
[Since I used to do that](https://github.com/eitsupi/r-ver/blob/8a3b1d4ee71f1c885d22df02041254c311ccecaf/dockerfiles/focal.Dockerfile#L26-L28), I thought it would be better not to compromise the portability of `install_R.sh` considering the possibility that there are other such users.

Another reason is that modifying `install_R.sh` would make debugging more difficult in terms of runtime. Why don't you run it in a separate layer like this PR first, and then integrate it into `install_R.sh` when the behavior is considered stable enough?

cboettig:
@eitsupi I'm with you 100%, we don't need to modify `install_R.sh`, sorry! I was just confused  by the additional changes to the cuda Dockerfiles because I forgot they now also run `install_R.sh` instead of inheriting from r-ver!  that's what I get for reviewing the PR from my phone.  this looks good.

eitsupi:
@cboettig Thank you for merging it.

Since #278 did not include the changes in this PR, a diff change (add `patch_install_command.sh` to the R4.1.2 images) was created another PR #279, so you seems need to merge that one as well.

PR #283, opened 2021-11-09
Install the duckdb package in `install_tidyverse.sh`

### LINKED ISSUE #267: Install the duckdb package with `install_tidyverse.sh` ?

eitsupi:
close #267
Note that this change assumes that the `install2.r` command has the `--skipmissing` option, which was added in #275.

Add the duckdb package, which can be used with dplyr and arrow.
We can read more about how to use it in the Apache Arrow blog post.
https://arrow.apache.org/blog/2021/11/08/r-6.0.0/

**ToDo**

- [x] Update NEWS.md

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,14 @@
+# 2021-11
+
+## Changes in rocker_scripts
+
+- `install_tidyverse.sh` now assumes that the `install2.r` command has the `--skipmissing` option. ([#283](https://github.com/rocker-org/rocker-versioned2/pull/283))
+- The [duckdb](https://github.com/duckdb/duckdb) package is added to the R packages installed by running `install_tidyverse.sh`. ([#283](https://github.com/rocker-org/rocker-versioned2/pull/283))
+
+## Changes in pre-built images
+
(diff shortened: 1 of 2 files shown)


eitsupi:
The error I get from checking the last Markdown file is "Multiple top level headers in the same document".
Is it better to uncheck this item or to change the header structure in `NEWS.md`?

cboettig:
seems reasonable that we change the structure of NEWS.md so it has a single h1 tag called NEWS or something, and other tags becoming subheadings?

PR #288, opened 2021-11-17
Some minor updates to Dockerfiles and install scripts

eitsupi:
I'm submitting this because I found several small improvements that have little impact on the behavior of the container.

1. CTAN mirrors are HTTPS enabled.
2. Some scripts are missing `set -e` and may succeed in the build even if the installation fails. Like that:  ml-verse_a68e3defd062  
It seems that the installation of wgrib2 failed because wget failed 20 retries, but the image was built as is and pushed to DockerHub.
3. Switch `libcurl4-gnutls-dev` to `libcurl4-openssl-dev` in `install_shiny_server.sh`, as described in https://github.com/rocker-org/rocker-versioned2/pull/156#issue-877264275.  
`rocker/shiny-verse` images has `libcurl4-openssl-dev` installed, but it doesn't seem to have any problems, so we can expect that replacing `libcurl4-gnutls-dev` in `install_shiny_server.sh` with `libcurl4-openssl-dev` will not cause any problems.

### DIFF
--- a/build/make-stacks.R
+++ b/build/make-stacks.R
@@ -96,8 +96,8 @@ library(gert)
 .latest_ctan_url <- function(date) {
   url <- dplyr::if_else(
     is.na(date),
-    "http://mirror.ctan.org/systems/texlive/tlnet",
-    stringr::str_c("http://www.texlive.info/tlnet-archive/", format(date, "%Y/%m/%d"), "/tlnet")
+    "https://mirror.ctan.org/systems/texlive/tlnet",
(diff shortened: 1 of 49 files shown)


cboettig:
this is great, thanks!  yes, definitely in favor of standardizing on `libcurl-openssl` flavor, I seem to recall some R packages don't like the gnutls version (though it is sometimes tricky to avoid in the deb dependencies, e.g. I think librdf0-dev pulls against that version and we swap it back out on verse)

PR #295, opened 2021-11-24
Proof-of-concept for code-server based interface

cboettig:
A proof-of-concept to enable users to swap in https://github.com/cdr/code-server as a possible alternative to RStudio Server.

Motivation: Rocker instances with RStudio built-in provide a natural/convenient way for users to interact with the container on a remote or local host through it's web-browser interface.  `code-server` provides a similar browser-based environment, only built upon Visual Studio instead.  In addition to just supporting the option of a different IDE interface that may be preferable to some users, the `code-server` approach may offer a few benefits to some of the long-standing challenges we've had, such as:

- running the container as non-root user (particularly for podman and other docker-based options that require this)
- built-in support for jupyter notebooks (our binder builds have some unresolved issues)
- allowing arbitrary number of tabs to be open to different (or the same) projects running in the same instance (no need to use different browsers etc)
- less coupling between the R session / active project and the editor and other IDE features (i.e. can restart the R session or switch projects without interrupting the active terminal, though maybe using a terminal multiplexer in RStudio is sufficient for that)
- some convenient integrations for existing OAuth authentication options, e.g. https://www.pomerium.io/guides/code-server.html

While this would provide  access to Visual Studio extensions, themes, and IDE features, note that code-server does not support the non-open-source extensions found in Visual Studio, such as LiveShare or CoPilot.

### DIFF
--- a/experimental/scripts/configure_vscode.sh
+++ b/experimental/scripts/configure_vscode.sh
@@ -1,0 +1,13 @@
+
+## Runs as user 
+
+install2.r languageserver lintr styler profvis httpgd
+sudo pip install radian
+
+## unfortunately these don't run and complete, but instead tries to bring up code-server
+#code-server install Ikuyadeu.r
+#code-server install RDebugger.r-debugger
+
+echo "options(vsc.use_httpgd = TRUE)" >> ~/.Rprofile
+
+
--- a/experimental/entrypoint.sh
+++ b/experimental/entrypoint.sh
@@ -1,0 +1,21 @@
+#!/bin/sh
+set -eu
+
+# We do this first to ensure sudo works below when renaming the user.
+# Otherwise the current container UID may not exist in the passwd database.
+eval "$(fixuid -q)"
+
+if [ "${DOCKER_USER-}" ]; then
+  USER="$DOCKER_USER"
+  if [ "$DOCKER_USER" != "$(whoami)" ]; then
+    echo "$DOCKER_USER ALL=(ALL) NOPASSWD:ALL" | sudo tee -a /etc/sudoers.d/nopasswd > /dev/null
+    # Unfortunately we cannot change $HOME as we cannot move any bind mounts
+    # nor can we bind mount $HOME into a new home as that requires a privileged container.
+    sudo usermod --login "$DOCKER_USER" coder
+    sudo groupmod -n "$DOCKER_USER" coder
+
+    sudo sed -i "/coder/d" /etc/sudoers.d/nopasswd
+  fi
+fi
+
+dumb-init /usr/bin/code-server "$@"
--- a/experimental/entrypoint.sh
+++ b/experimental/entrypoint.sh
@@ -11,10 +11,10 @@ if [ "${DOCKER_USER-}" ]; then
     echo "$DOCKER_USER ALL=(ALL) NOPASSWD:ALL" | sudo tee -a /etc/sudoers.d/nopasswd > /dev/null
     # Unfortunately we cannot change $HOME as we cannot move any bind mounts
     # nor can we bind mount $HOME into a new home as that requires a privileged container.
-    sudo usermod --login "$DOCKER_USER" coder
-    sudo groupmod -n "$DOCKER_USER" coder
+    sudo usermod --login "$DOCKER_USER" rocker
+    sudo groupmod -n "$DOCKER_USER" rocker
 
-    sudo sed -i "/coder/d" /etc/sudoers.d/nopasswd
+    sudo sed -i "/rocker/d" /etc/sudoers.d/nopasswd
   fi
 fi
 
--- a/experimental/scripts/install_vscode.sh
+++ b/experimental/scripts/install_vscode.sh
@@ -13,6 +13,7 @@ apt-get update \
     nano \
     git \
     procps \
+    pandoc \
     openssh-client \
     sudo \
     vim.tiny \
--- a/experimental/scripts/install_vscode.sh
+++ b/experimental/scripts/install_vscode.sh
@@ -2,25 +2,48 @@
 
 set -e
 
+apt-get update \
+ && apt-get install -y \
+    curl \
+    dumb-init \
+    zsh \
+    htop \
+    locales \
+    man \
+    nano \
+    git \
+    procps \
+    openssh-client \
+    sudo \
+    vim.tiny \
+    lsb-release \
+    tmux \
+  && rm -rf /var/lib/apt/lists/*
+
+DEFAULT_USER=${1:-${DEFAULT_USER:-rocker}}
+adduser --gecos '' --disabled-password ${DEFAULT_USER} && \
+  echo "${DEFAULT_USER} ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers.d/nopasswd
+addgroup ${DEFAULT_USER} staff
+
+## User name is always ${DEFAULT_USER} even when run with different $(id -u):$(id -g)
+## allow user name to be set with DOCKER_USER at runtime 
+ARCH="$(dpkg --print-architecture)" && \
+    curl -fsSL "https://github.com/boxboat/fixuid/releases/download/v0.5/fixuid-0.5-linux-$ARCH.tar.gz" | tar -C /usr/local/bin -xzf - && \
+    chown root:root /usr/local/bin/fixuid && \
+    chmod 4755 /usr/local/bin/fixuid && \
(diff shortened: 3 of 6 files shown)


experimental/scripts/configure_vscode.sh
```diff
@@ -0,0 +1,13 @@
+
+## Runs as user 
+
+install2.r languageserver lintr styler profvis httpgd
```
eitsupi, line 4: `lintr` and `styler` are basic dependencies of `languageserver`, so I don't think they need to be listed here.

experimental/scripts/configure_vscode.sh
```diff
@@ -0,0 +1,13 @@
...
+sudo pip install radian
+
+## unfortunately these don't run and complete, but instead tries to bring up code-server
+#code-server install Ikuyadeu.r
+#code-server install RDebugger.r-debugger
+
+echo "options(vsc.use_httpgd = TRUE)" >> ~/.Rprofile
```
eitsupi, line 11: The current vscode-R can describe this setting in `settings.json`, so it may be unnecessary.

eitsupi (commented):
If you want to add a script for vscode-R, you can refer to the definition I maintain for VSCode Remote-Containers or a script I wrote for `rocker/r-ver`.
https://github.com/microsoft/vscode-dev-containers/tree/main/containers/r
https://github.com/eitsupi/r-ver/blob/83e7f4d1be1bb8ec2ff1b3bfa8ce54c8f60fb3b1/scripts/install_editorsupports.sh

One of the advantages of vscode is that it can be run on arm64, but it also requires multiple apt packages for source installation.

I also think it would be useful to separate the contents of `install_rstudio.sh` into the true dependencies of RStudio Server and other useful settings that can be reused when installing editors other than RStudio Server.

experimental/scripts/configure_vscode.sh
cboettig: :+1:

experimental/scripts/configure_vscode.sh
cboettig: Ah, that's nice! I'm still very new to vscode-R -- where does that `settings.json` live?  Somewhere in the extensions dir presumably?  (as you can see in the note above, I haven't gotten `code-server` to install the extension automatically yet either...)

cboettig:
Thanks @eitsupi ! 

> If you want to add a script for vscode-R, you can refer to the definition I maintain for VSCode Remote-Containers or a script I wrote for rocker/r-ver.
https://github.com/microsoft/vscode-dev-containers/tree/main/containers/r
https://github.com/eitsupi/r-ver/blob/83e7f4d1be1bb8ec2ff1b3bfa8ce54c8f60fb3b1/scripts/install_editorsupports.sh

  I admit I hadn't seen these, looks fantastic.  Will have to give them a closer look. 
  
  > One of the advantages of vscode is that it can be run on arm64, but it also requires multiple apt packages for source installation.
  
  That's a good point, looks like the code-server version [should run on ARM](https://github.com/cdr/code-server/releases).  
  
  > I also think it would be useful to separate the contents of install_rstudio.sh into the true dependencies of RStudio Server and other useful settings that can be reused when installing editors other than RStudio Server.
  
  yup we should probably do that.  some of [those deps](https://github.com/rocker-org/rocker-versioned2/blob/master/scripts/install_rstudio.sh#L7-L23) are pretty universally useful and should probably just be re-listed in, say, `install_tidyverse.sh` and later scripts so they are installed if `rstudio` is skipped?   Others pretty specific to RStudio.

experimental/scripts/configure_vscode.sh
eitsupi: The user's settings.json needs to be stored in the directory that is created when vscode is installed.
https://code.visualstudio.com/docs/getstarted/settings

Typically, users have their own settings.json that they will sync in some way (Microsoft's Visual Studio Code has a proprietary sync solution built in that will sync to GitHub account or Microsoft account).
I only use VisualStudio Code, so I don't know how to reuse settings.json in code-server in general (perhaps bindmount?).
https://coder.com/docs/code-server/latest/install#docker

```shell
mkdir -p ~/.config
docker run -it --name code-server -p 127.0.0.1:8080:8080 \
  -v "$HOME/.config:/home/coder/.config" \
  -v "$PWD:/home/coder/project" \
  -u "$(id -u):$(id -g)" \
  -e "DOCKER_USER=$USER" \
  codercom/code-server:latest
```

yuvipanda:
@cboettig hopefully https://github.com/rocker-org/rocker-versioned2/pull/309 will fix the issues with the binder image.

benz0li:
@cboettig You may be interested in https://gitlab.b-data.ch/jupyterlab/r/docker-stack, which combines https://github.com/rocker-org/rocker-versioned with https://github.com/jupyter/docker-stacks including [code-server](https://github.com/coder/code-server) and further customisations.

See also https://demo.jupyter.b-data.ch for a minimal deployment (CPU: 2 cores, RAM: 4 GB) based on https://gitlab.com/b-data/docker/deployments/jupyter serving either [registry.gitlab.b-data.ch/jupyterlab/r/verse:latest-root](https://gitlab.b-data.ch/jupyterlab/r/verse/container_registry) or [registry.gitlab.b-data.ch/jupyterlab/julia/ver:latest-root](https://gitlab.b-data.ch/jupyterlab/julia/ver/container_registry).  
ℹ️ Both [JupyterHub](https://gitlab.b-data.ch/jupyterhub/jupyterhub/container_registry) and JupyterLab docker images are multi-arch (`linux/amd64`, `linux/arm64/v8`)

eitsupi:
@benz0li If you like, you can mention them on the recently renewed Rocker project web site?
I don't know how many people read this page, but I think it might help those interested in Docker images for R to know about them.
https://rocker-project.org/images/external.html

benz0li:
> @benz0li If you like, you can mention them on the recently renewed Rocker project web site?

Thank you for that suggestion. I will open a pull request at https://github.com/rocker-org/website in due time.

eitsupi:
Wondering if this could be closed in favor of https://github.com/rocker-org/devcontainer-images (devcontainer-base approach).
There are none at this time, but we can expect to see a code-server installation feature appear in the future.

cboettig:
@eitsupi is it possible to self-host a vscode instance / devcontainer that a user accesses in the browser over HTTPS (as in GitHub Codespaces) rather than over ssh tunnel?  My understanding is that vscode decided not to support that mechanism, and that was the main reason for https://coder.com/ 's fork of the project?   So I think the objectives here are somewhat distinct?

(though we still may not want to pursue any images with coder.com based interface anyway)

eitsupi:
@cboettig Rather than deciding not to support it, it may be closer to saying they did not OSS it.
The document of gitpod's openvscode-server, which is later than coder's code-server, mentions that.
https://github.com/gitpod-io/openvscode-server/#why

Microsoft has recently announced "the" VSCode Server that will run VSCode server on any computer and allow access over https.
I have not used it, but I think this is the solution you are looking for.
https://code.visualstudio.com/docs/remote/vscode-server
However, I have not used this and do not know if it is easy to integrate with devcontainer at this stage.

cboettig:
@eitsupi Cool, I hadn't seen the announcement for VSCode Server!

> The Visual Studio Code Server is a service you can run on a remote development machine, like your desktop PC or a virtual machine (VM). It allows you to securely connect to that remote machine from anywhere through a vscode.dev URL, without the requirement of SSH

That does seem like a nice analog to how we already package RStudio Server.  Looks like currently it's only available through a sign-up system so I guess we can wait and see if that becomes an open source piece of software we'd be able to package in a public/open docker container recipe or not....

Oh yes, & thanks for the gitpod link, that's exactly what I meant.  (obviously they already deploy vscode over https in proprietary systems like GitHub editor and GitHub codespaces).

cboettig:
hmm... also looks like the license may preclude such use https://code.visualstudio.com/docs/remote/vscode-server#_can-i-host-the-vs-code-server-as-a-service

eitsupi:
Since Microsoft products are intended to work with Microsoft services, we would need to install and use coder/code-server and gitpod-io/openvscode-server on the container in order to use them freely.
(As a side note, it appears that RStudio, PBC had an investment in openvscode-server)
https://www.gitpod.io/blog/openvscode-server-launch

eitsupi:
I think we can close this in favor of <https://github.com/rocker-org/devcontainer-features> and <https://github.com/loft-sh/devpod>.

cboettig:
agreed!

PR #298, opened 2021-11-29
Set a random password as default password of RStudio Server

### LINKED ISSUE #237: -e ROOT=true option not making rstudio user root
I'm starting rstudio server with

    docker run -d -p 8787:8787 -e ROOT=TRUE -e PASSWORD=rstudio rocker/rstudio

And yet

image

and

```

(base) balter@balter-host:~$ docker exec -it boring_buck bash
root@9cf43917d75b:/# su rstudio
rstudio@9cf43917d75b:/$ groups
rstudio staff

```

Also referencing https://github.com/rocker-org/rocker/issues/94#issuecomment-64753170

### LINKED ISSUE #294: Default password on rocker/rstudio
Related to #237

If we do not set the `-e PASSWORD=foobar` option when creating a new container with the `docker run` command, the `/init` command will exit halfway and RStudio Server will start without some option settings.
I think the current behavior makes little sense from a security point of view, because users may not notice the warning message and continue to use the system as it is, except when they try to set a specific option (like #237).

For example, Jupyter https://hub.docker.com/u/jupyter automatically generate a random token that the user can use to log in when started without the password setting.

If we create a container without the `-e PASSWORD=foobar` option, I think we can use the `pwgen` command to set a random password for the default user (rstudio).
Of course, if we make this change, we will also have to modify the existing documentation that says "the initial password is rstudio".

eitsupi:
Close #294, close #237

If the environment variable `PASSWORD` is not set, the RStudio user's password will be automatically generated and set when the `/init` command is executed.

image

**ToDo**

- [ ] Update docs.
  - [x] NEWS
  - [ ] The Rocker Project website https://www.rocker-project.org/use/managing_users/
  - [ ] DockerHub https://hub.docker.com/r/rocker/rstudio

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,11 @@
 # News
 
+## 2021-12
+
+### Changes in pre-built images
+
+- In the newly built images, the default password for RStudio Server will be randomly generated when the server is started. ([#298](https://github.com/rocker-org/rocker-versioned2/pull/298))
+
 ## 2021-11
 
 ### Changes in rocker_scripts
--- a/scripts/install_rstudio.sh
+++ b/scripts/install_rstudio.sh
@@ -19,6 +19,7 @@ apt-get install -y --no-install-recommends \
     psmisc \
     procps \
     python-setuptools \
+    pwgen \
     sudo \
     wget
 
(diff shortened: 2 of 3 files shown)


cboettig:
Excellent.  & I'll update the docker hub README and the website accordingly

PR #300, opened 2021-12-02
Causes automatic update workflows to fail if RSPM becomes unavailable

eitsupi:
Related to #299.

The workflow for automatically updating Dockerfile etc. communicates with [RStudio Public Package Manager](https://packagemanager.rstudio.com/) to explore available URLs.
If RSPM is not accessible, as it is today, the execution time of the workflow will be very long, so I will change the workflow to stop and error by setting a time limit.
And, add a badge to README.md so that we can check if there are any errors easily.

image


cboettig:
oh good call!  I like this approach.

Issue #191, opened 2021-06-25
Docker Hub rocker documentation 


stevegbrooks:
The docker hub site for rocker still mentions that the images are built on debian, but they're now being built on Ubuntu.

Should be an easy fix, and I'd submit the PR myself, but I just don't know where to do it.

Thanks

cboettig:
Thanks!  Can you link to the page you are seeing?  Docker Hub is a bit convoluted on this: pages "Overview" shows the README of the page that was once connected to an automated build. e.g. the overview you see at: https://hub.docker.com/r/rocker/rstudio comes from this readme: https://github.com/rocker-org/rocker-versioned/tree/master/rstudio#readme, even though the image is no longer built from that repo.  (Hub doesn't support enough tags in automated builds, and doesn't provide an independent way to re-wire which readme is displayed for which image!)

stevegbrooks:
https://hub.docker.com/r/rocker/verse

Here it mentions Debian:

"This repository provides alternate stack to r-base, with an emphasis on reproducibility. Compared to those images, this stack:

builds on debian stable (debian:jessie for versions < 3.4.0, debian:stretch after, etc) release instead of debian:testing, so no more apt-get breaking when debian:testing repos are updated and you had to muck with -t unstable to get apt-get to work."

eitsupi:
https://hub.docker.com/r/rocker/verse is linked to https://github.com/rocker-org/rocker-versioned, but the contents of README.md do not seem to be reflected.

DockerHub
![image](https://user-images.githubusercontent.com/50911393/129469831-8c895da3-6067-45f0-9f63-3a6dab21d8ed.png)

GitHub
![image](https://user-images.githubusercontent.com/50911393/129469837-f077464f-b0e1-4409-bd11-b8a3ca8372d6.png)

pgramme:
I also had the same problem: following the link from DockerHub to Github leads to the old repo (R v3.x), not here. I understand it would be best to have that DockerHub-GitHub link depending on the version but you can't.

Simple fix: Assuming most people are now interested in R 4.x, wouldn't it be useful to update the Github link on https://hub.docker.com/r/rocker/verse? (should point here)

dpprdan:
> hub.docker.com/r/rocker/verse is linked to rocker-org/rocker-versioned, but the contents of README.md do not seem to be reflected.

This is basically true for all versioned2 images, i.e. their README and their Source Repository links are outdated. 
[...]

PR #307, opened 2021-12-09
Add a new GitHub Actions workflow to update DockerHub descriptions

eitsupi:
Related to #191

Add a workflow to update the description on DockerHub with GitHub Actions.
It allows documents located in different repositories to be aggregated into this repository.
If you are able to consolidate the documentation, I recommend archiving the old GitHub repository to avoid confusion.

Since we don't know if it will work properly, I am first targeting only `rocker/cuda`, which is considered to be the least affected (since it was not used until a few months ago).

Current appearance https://hub.docker.com/r/rocker/cuda

image

### DIFF
--- a/.github/workflows/descriptions.yml
+++ b/.github/workflows/descriptions.yml
@@ -1,0 +1,29 @@
+name: Update Docker Hub Description
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - descriptions/*.md
+  workflow_dispatch:
+
+jobs:
+  dockerHubDescription:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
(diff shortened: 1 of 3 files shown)


cboettig:
Good idea, let's give this a try :rocket:

cboettig:
need to figure out where I am supposed to add credentials.  thought I already had those configured but https://github.com/rocker-org/rocker-versioned2/runs/4474499143?check_suite_focus=true#step:3:10

eitsupi:
This may be a limitation of the API.
As described in this issue (https://github.com/peter-evans/dockerhub-description/issues/10), it seems that you cannot use this API with an access token, and you have to set an actual password.

Is it possible to set a password? If so, please try to run the workflow manually after setting it.

PR #309, opened 2021-12-15
Require at least v2 of jupyter-rsession-proxy

yuvipanda:
Previous versions did not work with latest R or RStudio,
which is the primary reason the binder image hasn't been working
recently

The other reason the repo wasn't working was because `jupyter notebook`
was the start command used, but jupyter-rsession-proxy no longer depended
on the `notebook` package. I've fixed that by installing notebook and jupyterlab.


yuvipanda:
You can test this out in https://mybinder.org/v2/gh/yuvipanda/rocker-binder-template/HEAD?urlpath=rstudio. https://github.com/yuvipanda/rocker-binder-template/blob/0765b74a6af1bef418f425c8fa516d6e81d425ba/Dockerfile#L13 is the extra line I've added, equivalent to this PR

Issue #311, opened 2021-12-16
renv/CI focused images with geospatial sysdeps but no packages


noamross:
Hi all - a proposal for an additional package in the stack: For projects using `renv`, especially in the context of CI systems, it's not really needed to have lots of packages pre-installed as they will just be re-installed later.  What do we think of an image set that has all the _system_ dependencies up through `geospatial`, but no packages or RStudio installed?

eddelbuettel:
I like it. I do the same in the (smaller, more experimental) r-rspm image as "we know" what is missing and add it preemptively whereas this is and will always be the one sore spot and achilles heel of the RSPM packages.  

The only possible wrinkle I could think of is that it may interfere with the 'versioned2' ability to install at different time points.

cboettig:
seems reasonable to me!  looks like this could save > 1 GB in unpacked initial image size

eitsupi:
For this use, it may make more sense to install and manage R with rig.
https://github.com/r-lib/rig

PR #312, opened 2021-12-16
"CI" tag for system dependencies only

noamross:
A draft PR addressing #311, starting with just the script to install tidyverse/verse/geospatial system dependencies.

@eitsupi, I'm a bit confused as to where I would add a new image type to test.  `build/make-stacks.R` both reads from and writes to the JSON files in `stacks/`.  To add a new tag (I'm thinking `rocker/ci`), should I modify the stack files manually and then also add code to `build/make-stacks.R` for tagging?


scripts/install_renv_sysdeps.sh
```diff
@@ -0,0 +1,118 @@
...
+    qpdf \
+    texinfo \
+    software-properties-common \
+    vim
+
+# librdf0-dev depends on libcurl4-gnutils-dev instead of libcurl4-openssl-dev...
+# So: we can build the redland package bindings and then swap back to libcurl-openssl-dev... (ick)
+# explicitly install runtime library sub-deps of librdf0-dev so they are not auto-removed.
+apt-get install -y librdf0-dev
+install2.r --error --skipinstalled -n $NCPUS redland
```
noamross, line 65: @cboettig Not sure what is needed in this section

cboettig: yeah, this is a bit of a mess.  debian/ubuntu ship a few different versions of libcurl (specifically, gnutils vs openssl), which conflict.  (Some packages can bind either one, but don't necessarily play nicely with both -- e.g. I believe `curl` R package doesn't fail but does some unexpected things with the non-openssl version iirc?)  

The `redland` R package needs `librdf0-dev` which depends on gnutils, meaning that line 64 will uninstalls the openssl version and potentially breaks things.  So if we're going to install `librdf0-dev`, I *think* it is necessary to reinstall `libcurl4-openssl-dev` after we finish building `redland` from source (maybe not necessary now that we can get binaries??)

Anyway, long story short, I've always been very confused about the libcurl conflict.  @eddelbuettel or @jeroen probably have a much better handle on this; maybe one of them could advise what's the best way to support both redland and curl R packages?

eddelbuettel: Sorry, do not have a fix. There are alternative implementations, which in turn are use by different packages, so you could indeed end up with 'graph' with conflicts.  Best solution might be to locally rebuild packages using, say, gnutls to use openssl.  Not something I tried though.

eitsupi:
Is this image supposed to set a specific date snapshot as CRAN?
Or do you expect to use the "latest" for all R versions?

> I'm a bit confused as to where I would add a new image type to test. `build/make-stacks.R` both reads from and writes to the JSON files in `stacks/`. To add a new tag (I'm thinking `rocker/ci`), should I modify the stack files manually and then also add code to `build/make-stacks.R` for tagging?

Yes, you need to run the script after editing `stacks/devel.json` and `build/make-stacks.R` to update `stacks/4.1.2.json` and `stacks/4.1.1.json`.
Only the part of `stacks/devel.json` that is related to the R version is rewritten by the script to become `stacks/4.1.2.json` and `stacks/4.1.1.json`.

To update stack files with version <= 4.1.0, we will need to temporarily fix the version limit of the script and transcribe `devel.json`, or update them manually.
https://github.com/rocker-org/rocker-versioned2/blob/f3f9cafd25059f3b1e7d14e5d681f1f9063e4328/build/make-stacks.R#L540

If you can show me the structure of the Dockerfile you want, I can update the script for you.

eitsupi:
I think it would be better to set an option in each script to skip installation of R packages and call each script instead of copying the `apt install` command into this script to avoid code duplication.

PR #316, opened 2022-01-05
 Reports: Add Entrypoint and Cmd

eitsupi:
Since there are many Dockerfiles that do not contain `CMD` in this repository, it becomes a bit troublesome to check the `CMD` set in the container image.
So display the `Entrypoint` and `Cmd` of the container images so that these can be viewed on the wiki.

And, I did some minor refactoring.

### rocker/r-ver

r-ver

### rocker/binder

binder

### rocker/tidyverse

tidyverse


cboettig:
thanks, looks good!

Issue #282, opened 2021-11-06
Swich the base image to Ubuntu 22.04 LTS


eitsupi:
Just as the release dates of R4.0.0 and Ubuntu 20.04 were only a few days apart, the release date of R4.2.0 could be right after Ubuntu 22.04.
Immediately after the release of jammy, there is a possibility that RSPM binary installation, RStudio, etc. are not compatible with jammy, so we may need to modify the build system.

https://discourse.ubuntu.com/t/jammy-jellyfish-release-schedule/23906

```shell
$ docker run --rm -it rocker/r-ver Rscript -e "read.csv('/usr/share/distro-info/ubuntu.csv') |> tail(1)"
     version        codename series    created    release        eol eol.server
36 22.04 LTS Jammy Jellyfish  jammy 2021-10-14 2022-04-21 2027-04-21 2027-04-21
      eol.esm
36 2032-04-21
```

cboettig:
Yup.  Usually binaries compiled on the older system ought to run on the newer ones (though maybe Dirk can correct me on that); e.g. I believe in python world they build their 'universal' binary wheels by compiling with an ancient versions of linux.   So I'd think most packages would still work; though whether we want to ship any containers that install 20.04-based binaries on a 22.04 platform is another question.  iirc last time RSPM had a few month lag (and then there's the question of NVIDIA libraries, which I think were some 6 mo behind the release of 20.04 or so; though maybe different this time).

Of course, 20.04 has a lot more life in it.  I do think we want to stick with the rocker-versioned tradition of running with the latest stable release, but it does raise the issues of precisely when we switch, and if we continue to support 20.04-flavored images.  With the new automation and new buildkit approach to tags, I imagine that is somewhat easier?  Providing the same stack on multiple base distros seems pretty standard for docker projects these days.  Thoughts?

eddelbuettel:
> Usually binaries compiled on the older system ought to run on the newer ones (though maybe Dirk can correct me on that)

Yes they do, almost always (more below).  For example, over in PPA land for example Michael prepares the 5000+ packages in his repo on the LTS cadence, and I while I generally prefer CRAN-from-source I have one laptop where I use our lovely BSPM setup with the c2d4u PPA and it is awesome and generally works.

[...]

PR #319, opened 2022-01-07
Improvement of automatic stack files update system

eitsupi:
Some improvements have been made.

- No longer necessary to install development version of the `pak` package.
- Removed meaningless fallback when setting up a CRAN mirror and raise an error if all repositories are not available. (suggested by https://github.com/rocker-org/rocker-versioned2/pull/299#issuecomment-984845516)
- Automatically set the base image of `rocker/cuda:cuda11.1` (e.g. `nvidia/cuda:11.1.1-cudnn8-devel-ubuntu20.04`), it related to #282.
- Minor refactoring.

### DIFF
--- a/.github/workflows/dockerfiles.yml
+++ b/.github/workflows/dockerfiles.yml
@@ -15,7 +15,7 @@ jobs:
       - uses: actions/checkout@v2
       - name: install packages
         run: |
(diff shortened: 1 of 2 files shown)


cboettig:
Nice, all looks good to me!

PR #331, opened 2022-01-11
Add xgboost stan install and test scripts

RobJY:
Here's code to install and test gpu supported xgboost and stan.  Please let me know if I should move or add anything.

I'm not sure if I've put `install_stan.R` in the right place since it needs to be run by users to install stan.  Is there a better place for it.

I got the xgboost test script from `https://rdrr.io/cran/xgboost/src/demo/gpu_accelerated.R`.  Please let me know if you think there's something better we could use.

### DIFF
--- a/scripts/install_stan.R
+++ b/scripts/install_stan.R
@@ -1,0 +1,6 @@
+# install stan
+# users should run this script to install stan their R session
+
+install.packages("cmdstanr", repos = c("https://mc-stan.org/r-packages/", getOption("repos")))
+library(cmdstanr)
+install_cmdstan()
--- a/scripts/install_xgboost.sh
+++ b/scripts/install_xgboost.sh
@@ -1,0 +1,38 @@
+# Xgboost
+
+set -e
+set -x
+
+apt update && apt install -y software-properties-common
+
+# apt-get update
+# apt-get install -y --allow-downgrades --allow-remove-essential --allow-change-held-packages --allow-unauthenticated --no-install-recommends --no-upgrade \
+#   gcc-8 \
+#   g++-8 \
+#   libnccl2 \
+#   libnccl-dev
+
+# install nvidia toolkit
+apt-key adv --fetch-keys https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/7fa2af80.pub
+add-apt-repository "deb https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/ /"
+apt update && apt install -y nvidia-cuda-dev nvidia-cuda-toolkit
+
+# install cmake
+apt update && apt install -y build-essential libssl-dev gcc-8 g++-8
+update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-8 80 --slave /usr/bin/g++ g++ /usr/bin/g++-8 --slave /usr/bin/gcov gcov /usr/bin/gcov-8
+mkdir /tmp/cmake
+cd /tmp/cmake
+wget https://github.com/Kitware/CMake/releases/download/v3.20.0/cmake-3.20.0.tar.gz
+tar -zxvf cmake-3.20.0.tar.gz
+cd /tmp/cmake/cmake-3.20.0
+./bootstrap
+make
+make install
+
+# compile xgboost
+cd /root
+git clone --recursive https://github.com/dmlc/xgboost
+mkdir /root/xgboost/build
+cd /root/xgboost/build
+cmake .. -DUSE_CUDA=ON -DR_LIB=ON -DUSE_NCCL=ON -DNCCL_ROOT=/usr/lib/x86_64-linux-gnu
+make install -j4
--- a/scripts/tests/stan.R
+++ b/scripts/tests/stan.R
@@ -1,0 +1,12 @@
+library(cmdstanr)
+# Generate some fake data
+n <- 250000
+k <- 20
+X <- matrix(rnorm(n * k), ncol = k)
+y <- rbinom(n, size = 1, prob = plogis(3 * X[,1] - 2 * X[,2] + 1))
+mdata <- list(k = k, n = n, y = y, X = X)
+download.file("https://raw.githubusercontent.com/stan-dev/cmdstanr/master/vignettes/articles-online-only/opencl-files/bernoulli_logit_glm.stan", destfile = "test.stan")
+mod_cl <- cmdstan_model("test.stan",
+                        cpp_options = list(stan_opencl = TRUE))
+fit_cl <- mod_cl$sample(data = mdata, chains = 4, parallel_chains = 4,
+                        opencl_ids = c(0, 0), refresh = 0)
--- a/scripts/tests/xgboost.R
+++ b/scripts/tests/xgboost.R
@@ -1,0 +1,45 @@
+# An example of using GPU-accelerated tree building algorithms
+#
+# NOTE: it can only run if you have a CUDA-enable GPU and the package was
+#       specially compiled with GPU support.
+#
+# For the current functionality, see
+# https://xgboost.readthedocs.io/en/latest/gpu/index.html
+#
+
+library('xgboost')
+
+# Simulate N x p random matrix with some binomial response dependent on pp columns
+set.seed(111)
+N <- 1000000
+p <- 50
+pp <- 25
+X <- matrix(runif(N * p), ncol = p)
+betas <- 2 * runif(pp) - 1
+sel <- sort(sample(p, pp))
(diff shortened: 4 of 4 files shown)


cboettig:
These look good to me! @noamross any thoughts here?

noamross:
Only that I'm not sure these are updated since @RobJY found the issue in #342, so these should likely change to install something other than `nvidia-cuda-toolkit`, right @RobJY?

noamross:
Also, these are interactive test scripts, not scripts that return a value or a specific error for failure. Wouldn't it be better to have something that can integrate into a smoke test?

RobJY:
That's right @noamross, I'll have to update the xgboost install script when I've got a fix working.

scripts/install_xgboost.sh
```diff
@@ -0,0 +1,38 @@
+# Xgboost
```
eitsupi, line 1: Sheban is not written, is it correct?

scripts/install_stan.R
```diff
@@ -0,0 +1,6 @@
+# install stan
+# users should run this script to install stan their R session
+
+install.packages("cmdstanr", repos = c("https://mc-stan.org/r-packages/", getOption("repos")))
+library(cmdstanr)
+install_cmdstan()
```
eitsupi, line 6: Does this file need to be an R script? The other script files are shell scripts.

scripts/install_xgboost.sh
```diff
@@ -0,0 +1,38 @@
...
+#   libnccl2 \
+#   libnccl-dev
+
+# install nvidia toolkit
+apt-key adv --fetch-keys https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/7fa2af80.pub
+add-apt-repository "deb https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/ /"
+apt update && apt install -y nvidia-cuda-dev nvidia-cuda-toolkit
+
+# install cmake
+apt update && apt install -y build-essential libssl-dev gcc-8 g++-8
```
eitsupi, line 21: The other script deletes the cache after running apt update. Is this not necessary?

RobJY: You're right, that second `apt update` in not needed.  I'll remove it in my next commit.  Thanks @eitsupi!

scripts/install_xgboost.sh
RobJY: You're right @eitsupi I should have a shebang here.  Just added one in the last commit.  Thanks!

scripts/install_stan.R
RobJY: Good point @eitsupi.  I've changed it to a shell script.  Thanks!

RobJY:
I've pushed an update based on a fix that I'll be deploying to our systems tomorrow.  Let me explain what I think the issue is and a possible solution.

The issue we've been having seems to be that because the cuda libraries are being so actively developed it's very easy for a library to be updated after the cuda driver has been installed and then the library versions on the host and in the container no longer match.  Keeping the `nvidia-cuda-toolkit` would probably be fine, but it contains so many packages that it's more likely to run into this problem.  For this reason, the updated code uses a minimal set of libraries to help mitigate the issue.
Updating the driver would require scheduling a downtime for machine reboot, so it's better for us to lock the library versions and use `apt-mark hold` to try to keep them from being updated.  This has worked in my test container.

It seems like using this script for Rocker builds has the opposite problem in that we will be building images with the current versions of the libraries, but will have no way of knowing what version of the driver users will have on their host machines.  It seems very likely that this will often result in this library mismatch error.  For this reason I removed the versions from the `apt` call and now that I think of it we probably don't want to hold the installed version either.  I'll update that too.  Maybe the best way to deal with this is to have a document that helps users performs a set of steps to try to resolve the problem if they receive the library mismatch error message.  That may be as simple as updating their driver and then updating the libraries in the container.

Is there a better way to handle this?

RobJY:
> Also, these are interactive test scripts, not scripts that return a value or a specific error for failure. Wouldn't it be better to have something that can integrate into a smoke test?

I was thinking we'd collect standard out from running this and get the return code.  If the code isn't 0 we would just echo the standard out.  I see now I should do that inside the script though and not create more work outside it.
Thanks for pointing this out @noamross!

I've added the bash scripts that do what I described above.  Let me know if you'd like any other changes.

RobJY:
I just took another look at this to see if I could move it along.  I could add some text about possible fixes a user could try if they encounter the `Failed to initialize NVML: Driver/library version mismatch` error message, but I think we first need to learn a little more about the issue on a long running server like we have on premises with the fixes we've proposed there.  Those being that we will not peg versions of GPU packages and will install a new version of the Nvidia driver with each update.  This should cut down on the number of this kind of mismatch errors, but I'm not sure it will completely eliminate them.  I'll update this PR when we've gone through another local update and deploy.

cboettig:
@RobJY thanks for the follow-up.  the driver mismatch error is definitely an annoying one and not one that I've completely wrapped my head around either.  A few comments/observations so far:

>  Curiously, nvidia-smi shows CUDA version = 11.7 on both the host and in the container after running the script.

This is as expected.  `nvidia-smi` shows the CUDA driver version, i.e. the host version, not the version of the toolkit software.  The driver version must be >= the toolkit version accoding to NVIDIA.  

Updating the driver on the host while the container is still running is one way to get the `Failed to initialize NVML: Driver/library version mismatch`, but I also see this error when I don't think I've done any updates on the host?  Have you observed that?  Either way, simply doing `docker restart <container_id>` resolves the error, though obviously not ideal.

RobJY:
Thank you @cboettig for the explanation of the CUDA version from nvidia-smi!  That makes sense.

You're right about suggesting a machine reboot.  That should be the first thing we ask users to try.  I did see that that was successful for many users on the forums.  Unfortunately, it didn't work for me.

Let me explain what I think the issue was for our servers and  please let me know if you think I've got any of it wrong.
At the top of my attached script we install the updated driver and container toolkit.  I think this installs package version greater than those used by CUDA 11.1.  To counteract this I install the 11.1 toolkit on the host machine at the bottom of the script.  This downgrades the versions previously installed.
I have two pieces of evidence that this is what's happening:
1) I no longer get the package mismatch error
2) It fails if I don't use the flag `--allow-downgrades` in my call to `apt` at the top of the script with a message about not being able to satisfy dependencies when installing CUDA toolkit 11.1.

cboettig:
Thanks @RobJY , apologies I missed that the script was not part of the commit but just attached to the comment.  Very interesting! if it works it works, but honestly this just deepens the puzzle for me -- this script is all being run outside the container, on the host machine, yeah?  

I don't think you should need the toolkit of any kind installed on the host machine though?  Maybe I'm just confused.  All software above the kernel level (i.e. not including the drivers) should be handled in the containerized layer?  Or does that assumption go out the window when we're using nvidia-docker?  

I've also had to mess with apt-pinning to get nvidia container runtime to play nicely with upgraded drivers, as described here: https://github.com/pop-os/pop/issues/1708#issuecomment-1003422245 .  Maybe this is related to / an alternative version of your script above?

But I still see containers that have been up a while start to throw the error `Failed to initialize NVML: Unknown Error` whenever i try GPU operations (e.g. `nvidia-smi`) until I restart them, even when (I think) I haven't updated the drivers, so I don't really know why that is...  in these cases, the host can still run `nvidia-smi` no problem, it's just the container that is stuck.  Is that the situation you were seeing as well?

RobJY:
Apologies about the attached script.  I did have it hidden at the bottom of the post.  Yes, we're running the attached script on the host machine.

Yeah, I think you're right that the nvidia-docker install loads some packages on the host and those are at a higher version than those in the container.  This makes sense to me because the without any pinning it's going to install the newest versions available.  I also tried pinning versions earlier, but Nvidia doesn't seem to distribute older versions of packages for very long and when I went back to use the pinned script I had the older versions of the packages were unavailable.

I haven't seen the `Failed to initialize NVML: Unknown Error` message before.  I wonder if something running on the GPU got the card(s) into a bad state.  I was just experimenting with some ML code that can make use of multiple cards.  We have 2 cards of one type and 2 cards of another type on the machine.  When I tried running the code with one or two cards it was fine, but when I tried 3 or 4 it crashed and in such a way that we weren't able to use those cards until the machine was rebooted.  I saw behavior like this at a previous job where we were doing GPU development.  When a card crashes it seems to take a reboot to make it usable again.  I guess it makes sense that it can't report the nature of the error in that case, just that the card is not responsive.

RobJY:
That's interesting.  It looks like a nice clean solution to me.  It probably protects you from other potential issues as well.

PR #335, opened 2022-01-17
Add `distill` as package

joundso:
I think [`distill`](https://rstudio.github.io/distill/) would be a great candidate to be integrated in rocker/verse as it says to provide "(...) publishing-related packages".


cboettig:
agree, seems a natural fit here

PR #334, opened 2022-01-14
Add `Versions.md` to the wiki

eitsupi:
Related to #201

Now that #328 has been merged, we can generate a table to see the list of configuration values for each version, like https://github.com/rocker-org/rocker-versioned/blob/master/VERSIONS.md
I think it would be useful if this could be registered in the wiki and automatically updated.

### Versions.md

image

### sidebar

image

**ToDo**

- [x] Add sidebar of the wiki
- [x] Update the job to update wiki
- [x] Update `build/README.md` (About the disappearance of the corresponding version from `Versions.md` when the stackfile is deleted)

Issue #201, opened 2021-07-28
Are versioned rocker images always guaranteed to contain the same R package versions?


myoung3:
I was previously under the impression that the most recent version-tagged rocker image (e.g.  currently geospatial:4.1.0) is not stable with respect to R package versions. I got this (mistaken?) idea from this quote:

> Note that the MRAN date associated with the current release (e.g., 3.4.2 at the time of writing) will continue to advance on the Docker-hub image until the next R release.
https://journal.r-project.org/archive/2017/RJ-2017-065/RJ-2017-065.pdf

I was interpreting this quote to mean that the *most recent* version-tagged rocker image (e.g. 4.1.0) might have its installed R package versions change between the day 4.1.0 was released on rocker through the day the next version is released on rocker (i.e. 4.1.1) because the built-in MRAN date stored in 4.1.0 can change as 4.1.0 is updated during that period. 

I now realize that rocker-versioned2 no longer uses MRAN so that quote doesn't even apply anymore.  Nevertheless, I'm still uncertain of how package version freezing/pinning works for the versioned images. It would be great if there were documentation on this (maybe in the readme?). Some documentation on the following specific topics would be great:

- What builds (e.g. versioned vs latest) have stable, pinned R packages. The readme mentions that rstudio package manager is used for installing packages, but the readme doesn't even mention anywhere that R package versions are guaranteed to be constant over time within a specific versioned rocker image. And it would be a good idea to especially call out whether the most recent versioned rocker images are stable over time with respect to R package versions, given the above quote that confused me.
- How exactly package versions are pinned. It would be great to give a general description of how rstudio package manager is utilized to pin versions (ie what is the process for determining which package versions end up in a versioned rocker image)?
- Recommendations/best practices for how users might extend the rocker images by building on top of rocker images with even more R packages (with pinned version) in a way that's reproducible and consistent with the how the packages are specified in the versioned rocker images (ie should we use `devtools::install_version`? Or some other approach?)

Posting this issue on @eddelbuettel's recommendation after I said some things about rocker on twitter that made it apparent I was confused about how this all worked. 

Thanks!

cboettig:
@myoung3 Apologies for the confusion.

[...]

PR #347, opened 2022-02-13
[Docs] Rewrite the README

eitsupi:
Related to #201 and #334.

It's been about two years since this repository was created (or R4.0.0 was released), and I think it needs to be rewritten so that more of the content can be understood in the README for this repository alone, rather than in comparison to older repositories.
And, I also thought I should add some explanations about the build system and release cycle and the `devel` tag.

@cboettig @noamross I have rewritten most of the content. Could you please review it?


cboettig:
Thanks, looks great!

PR #368, opened 2022-02-27
Support for quarto and pandoc bundled with RStudio

### LINKED ISSUE #367: RStudio v2022.02.0+443 includes quarto instead of pandoc
```shell
$ ls /usr/lib/rstudio-server/bin/
crash-handler-proxy  crashpad_http_upload  postback  r-ldpath   rserver      rsession
crashpad_handler     pam-helper            quarto    rpostback  rserver-pam  rstudio-server
```

The pandoc executable is contained in the quarto directory, so it does not exist in the path referenced in the script below.

https://github.com/rocker-org/rocker-versioned2/blob/e47fcb9e69c8d90d79f216be8aaee9b626ab693b/scripts/install_pandoc.sh#L21

It might also be useful to add a new option to the script that installs quarto that uses the quarto included in RStudio. (Currently, quarto is updated frequently, so the default behavior is to install the latest version, which makes sense.)

eitsupi:
Fix #367

The latest release version of RStudio (2022.02.0+443) now bundles quarto instead of pandoc, which was previously bundled.
Therefore, the pandoc path bundled with RStudio is different from the path hard-coded in `install_pandoc.sh`.

This PR fixes this issue and makes it compatible with the new bundled version of pandoc path.
And adds an option to `install_quarto.sh` to use the RStudio bundled version of quarto.

I also noticed that the `pandoc-citeproc` executable files that once existed are no longer present.

```shell
$ docker run --rm -it rocker/rstudio:4.1.1 bash
root@30c15e2ae440:/# ls -l /usr/local/bin/pandoc*
lrwxrwxrwx 1 root root 41 Nov  1 18:41 /usr/local/bin/pandoc -> /usr/lib/rstudio-server/bin/pandoc/pandoc
lrwxrwxrwx 1 root root 50 Nov  1 18:41 /usr/local/bin/pandoc-citeproc -> /usr/lib/rstudio-server/bin/pandoc/pandoc-citeproc
root@30c15e2ae440:/# ls /usr/lib/rstudio-server/bin/pandoc/pandoc*
/usr/lib/rstudio-server/bin/pandoc/pandoc
```

I changed so that symbolic links will not be created if this executable does not exist.

### DIFF
--- a/scripts/install_pandoc.sh
+++ b/scripts/install_pandoc.sh
@@ -40,7 +40,9 @@ if [ "$PANDOC_VERSION" != "$INSTALLED_PANDOC_VERSION" ]; then
 
   if [ "$PANDOC_VERSION" = "$BUNDLED_PANDOC_VERSION" ] || [ "$PANDOC_VERSION" = "default" ]; then
     ln -fs "$BUNDLED_PANDOC" /usr/local/bin
-    ln -fs "${BUNDLED_PANDOC}-citeproc" /usr/local/bin
+    if [ -f "${BUNDLED_PANDOC}-citeproc" ]; then
+      ln -fs "${BUNDLED_PANDOC}-citeproc" /usr/local/bin
+    fi
   else
     if [ -L "/usr/local/bin/pandoc" ]; then
       unlink /usr/local/bin/pandoc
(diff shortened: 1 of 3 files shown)


cboettig:
:+1: Thanks, LGTM!

PR #381, opened 2022-03-09
Add prefix "R-" to the `org.opencontainers.image.version` annotation

eitsupi:
Add prefix `R-` to `org.opencontainers.image.version`, which represents the major software included in the image, whereas up to now it has listed just the version without prefix.
https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys

For example, the content of the `org.opencontainers.image.version` annotation of `tonistiigi/binfmt:latest`, which used for multi-platform builds in GitHub Actions, is `qemu-v6.1.0-21`.
https://github.com/rocker-org/rocker-versioned2/runs/5355027888?check_suite_focus=true#step:4:172

### DIFF
--- a/bakefiles/4.0.0.docker-bake.json
+++ b/bakefiles/4.0.0.docker-bake.json
@@ -39,7 +39,7 @@
         "org.opencontainers.image.title": "rocker/r-ver",
         "org.opencontainers.image.description": "Reproducible builds to fixed version of R",
         "org.opencontainers.image.base.name": "docker.io/library/ubuntu:focal",
-        "org.opencontainers.image.version": "4.0.0"
+        "org.opencontainers.image.version": "R-4.0.0"
       },
       "tags": [
         "docker.io/rocker/r-ver:4.0.0"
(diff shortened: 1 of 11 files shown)


eddelbuettel:
Hm, I find 'qemu-v6.1.0-21' more readable than 'R4.0.2'.  Could we maybe use 'R-4.0.2' or 'R 4.0.2' or 'R (4.0.2)' or 'R-v4.0.2' ?
My personal vote would be 'R-4.0.2' as spaces may be outlawed, and I find lowercase-v to be redundant in versions.

eitsupi:
Thank you for sharing your opinion.
I agree, I thought about something like `R-version-4.1.2` that appears when R is started but it is definitely redundant so `R-4.1.2` looks better.

eddelbuettel:
Thanks for the quick reply!  (Timezones can help sometimes ...)   R is super confusing here as for historical (?) reason there are nearly half a dozen ways to get at it.  But it is a clean "GNU-style" major.minor.patch so showing that is good.

PR #391, opened 2022-03-20
[Docs] Updated to reflect `Versions.md` in the wiki

eitsupi:
Continuation of the work in #334.

1. Add a note about version fixing and a link to [Versions.md](https://github.com/rocker-org/rocker-versioned2/wiki/Versions) in the README
2. [As noted here](https://github.com/rocker-org/rocker-versioned2/blob/935967b2bfb6581e11630f02d9f9f587031cad5d/build/README.md?plain=1#L26-L30), if you deprecate a particular stack file, do not move the stack file.
  This is because Versions.md refers to stack files.
  The `deprecated-stacks` directory is no longer needed due to this change in operation, so it is deleted.


cboettig:
great!

PR #395, opened 2022-03-26
Refactoring core Rocker scripts

eitsupi:
Related to #323


cboettig:
re-linted code all looks great here.  :tada:

eitsupi:
It appears that the post-merge build failed, but this was due to a network malfunction.

PR #398, opened 2022-04-01
Switch to GitHub's auto generating release notes

eitsupi:
GitHub now generates nice-looking release notes by default, so change the workflow to use it.
https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes

This change simplifies the workflow and generates release notes with the same structure commonly used on GitHub.

**ToDo**

- [x] Workflow test.


eitsupi:
I have tried it on my fork and this workflow seems to work correctly. (although the text is mostly empty because there was no PR on my fork)

image

PR #400, opened 2022-04-01
Rename bionic image definition files and improve test workflow for docker-bake.json

### LINKED ISSUE #399: `buildx bake` command not allow dots in target names (buildx v0.8)
Dots in the target name is no longer allowed in the latest release of buildx (docker/buildx#929).
If the target name in a `docker-bake.json` file contains a dot, the following error will occur when executing the `bake --print` command.

> Invalid name; only "[a-zA-Z0-9_-]+" are allowed

Currently some target names contain dots, which need to be changed.

https://github.com/rocker-org/rocker-versioned2/blob/98738ed87f37ff1e469fd84364f7ef4efc1cc241/bakefiles/4.0.0.docker-bake.json#L281

eitsupi:
Close #399
And, the test for `docker-bake.json` files is not needed for most Pull Requests, so it is separated into a single workflow file.

Note that this PR only changes the name used in the definition files and does not affect the images to be built.


cboettig:
nice solution, this looks good to me, merge when ready!

Issue #390, opened 2022-03-16
fixing the BLAS hot-switching mechanism


kforner:
There is a regression with the BLAS switching mechanism, as described here: https://www.rocker-project.org/use/other_topics/, 
in rocker-versioned2:

```shell
%docker run --rm rocker/verse:4.1.2 bash

R -q -e 'sessionInfo()' | grep so
BLAS/LAPACK: /usr/lib/x86_64-linux-gnu/openblas-pthread/libopenblasp-r0.3.8.so

update-alternatives --set libblas.so.3-x86_64-linux-gnu /usr/lib/x86_64-linux-gnu/blas/libblas.so.3

# nothing changed
R -q -e 'sessionInfo()' | grep so
BLAS/LAPACK: /usr/lib/x86_64-linux-gnu/openblas-pthread/libopenblasp-r0.3.8.so

# expected considering that
# ldd /usr/local/lib/R/lib/libR.so | grep blas
	libopenblas.so.0 => /lib/x86_64-linux-gnu/libopenblas.so.0 (0x00007f20400b0000)


```

This seems to be related to the auto-detection of the openblas lib during the R configuration, so that the `--with-blas="-lopenblas"` is used (cf https://cran.r-project.org/doc/manuals/r-devel/R-admin.html#OpenBLAS).

One simple way to fix that is to postpone the openblas installation AFTER the R compilation:

```diff
git diff scripts/install_R.sh
diff --git i/scripts/install_R.sh w/scripts/install_R.sh
index bad57c4..30f996b 100755
--- i/scripts/install_R.sh
+++ w/scripts/install_R.sh
@@ -24,11 +24,7 @@ export DEBIAN_FRONTEND=noninteractive
 R_HOME=${R_HOME:-/usr/local/lib/R}
 
 READLINE_VERSION=8
-OPENBLAS=libopenblas-dev
-if [ ${UBUNTU_VERSION} == "bionic" ]; then
-  READLINE_VERSION=7
-  OPENBLAS=libopenblas-dev
-fi
+
 
 apt-get update \
   && apt-get install -y --no-install-recommends \
@@ -46,7 +42,6 @@ apt-get update \
     libicu* \
     libpcre2* \
     libjpeg-turbo* \
-    ${OPENBLAS} \
     libpangocairo-* \
     libpng16* \
     libreadline${READLINE_VERSION} \
@@ -127,6 +122,14 @@ make
 make install
 make clean
 
[...]

PR #386, opened 2022-03-14
Split `install_R.sh` into two scripts and some improvements in the R installation process

eitsupi:
Close #385
Close #390

**ToDo**

- [x] Add `setup_R.sh` and move some lines from `install_R.sh` to `setup_R.sh`
- [x] Add a new GitHub Actions workflow to test building `rocker/r-ver`
- [x] Rename `install_R.sh` to `install_R_source.sh`
- [x] Merge `patch_install_command.sh` into `setup_R.sh`
- [x] Refactor `setup_R.sh`
- [x] Update Dockerfiles
- [x] Change BLAS installation order (#390)
- [x] Download R from the 0-Cloud CRAN mirror.
- [x] Update NEWS

PR #412, opened 2022-04-11
Remove `liblapack-dev` after R installation to reduce image size

eitsupi:
Related to #390

In #386, the image size increased by 30MB as a result of the additional installation of `liblapack-dev`.
However, this appears to be acceptable to remove after R installation, so I would like to remove it and reduce the image size.

### DIFF
--- a/scripts/install_R_source.sh
+++ b/scripts/install_R_source.sh
@@ -43,7 +43,6 @@ apt-get install -y --no-install-recommends \
     libbz2-* \
     libcurl4 \
     libicu* \
-    liblapack-dev \
     libpcre2* \
     libjpeg-turbo* \
     libpangocairo-* \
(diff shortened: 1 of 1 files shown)


kforner:
FWIW, the problem seems to lie with `/usr/local/lib/R/modules/lapack.so`:
```
$ ldd /usr/local/lib/R/modules/lapack.so 
	linux-vdso.so.1 (0x00007ffed91ff000)
	libR.so => not found
	liblapack.so.3 => not found
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fcebcc6e000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fcebca7c000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fcebcdd6000)

```

Issue #302, opened 2021-12-03
"cannot change locale" warnings


dpprdan:
I get "cannot change locale" warnings, when I build r-ver (4.1.2) images. They occur on GHA here as well: 

E.g. [here](https://github.com/rocker-org/rocker-versioned2/runs/4355910963?check_suite_focus=true#step:6:114)

```sh
/bin/bash: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)
```

and [here](https://github.com/rocker-org/rocker-versioned2/runs/4355910963?check_suite_focus=true#step:6:186)

```sh
#10 6.729 perl: warning: Setting locale failed.
#10 6.729 perl: warning: Please check that your locale settings:
#10 6.729 	LANGUAGE = (unset),
#10 6.729 	LC_ALL = "en_US.UTF-8",
#10 6.730 	LANG = "en_US.UTF-8"
#10 6.730     are supported and installed on your system.
#10 6.730 perl: warning: Falling back to the standard locale ("C").
```

I can't really tell whether it succeeds in the end. I just noticed that the following code [is recommended](https://github.com/docker-library/docs/tree/master/ubuntu#locales) for the official Ubuntu image:

```sh
RUN apt-get update && apt-get install -y locales && rm -rf /var/lib/apt/lists/* \
	&& localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8
ENV LANG en_US.utf8
```

cboettig:
Thanks @dpprdan .  We do install locales and configure as such:

https://github.com/rocker-org/rocker-versioned2/blob/master/scripts/install_R.sh#L59-L61

but maybe we ought to switch to that recommended way using  `localedef` instead?  

We also set LANG, https://github.com/rocker-org/rocker-versioned2/blob/master/dockerfiles/r-ver_4.1.2.Dockerfile#L11, but maybe that needs to be `utf8` instead of `UTF-8`?  I'm not really up-to-speed on locale issues, @eddelbuettel probably knows more about this.  Dirk, would you recommend we set locales as Daniel recommends above instead?  Or is something else the issue here perhaps?

eddelbuettel:
[...]

PR #413, opened 2022-04-12
Fix default locale settings

### LINKED ISSUE #397: Encoding problem shiny app on Shiny Server
Dear rockers,
I have some issues with showing specific characters in a dashboard.
An exactly the same dashboard was hosted with shiny-verse 4.0.5 and with shiny-verse 4.1.3.

In the 4.0.5 version this text is shown: beïnvloeden
In the 4.1.3 version this text is shown: be<c3><af>nvloeden
(all files are in UTF-8)

Any clue what it can be? Using docker-compose exec, I checked if the environment encoding and that seems okay (LANG=en_US.UTF-8).

eitsupi:
Fix #397

The change in #313 results in the deletion of the `update-locale` process, which causes the problem with Shiny Server.
This PR will re-correct how the locale is set up.

Note: The LANG environment variable is set after the execution of `install_R_source.sh`, since setting the LANG environment variable before installing the `locales` package would result in a warning (as described in https://github.com/rocker-org/rocker-versioned2/issues/302#issuecomment-997437630).

### DIFF
--- a/scripts/install_R_source.sh
+++ b/scripts/install_R_source.sh
@@ -15,9 +15,10 @@ R_VERSION=${1:-${R_VERSION:-"latest"}}
 apt-get update
 apt-get -y install locales lsb-release
 
-## Configure default locale, see https://github.com/docker-library/docs/tree/master/ubuntu#locales
-localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8
+## Configure default locale
 LANG=${LANG:-"en_US.UTF-8"}
+/usr/sbin/locale-gen --lang "${LANG}"
+/usr/sbin/update-locale --reset LANG="${LANG}"
 
 UBUNTU_VERSION=$(lsb_release -sc)
 
--- a/scripts/install_R_source.sh
+++ b/scripts/install_R_source.sh
@@ -6,7 +6,7 @@
 ## ex. latest, devel, patched, 4.0.0
 ##
 ## 'devel' means the prerelease development version (Latest daily snapshot of development version).
-## 'pached' means the prerelease patched version (Latest daily snapshot of pached version).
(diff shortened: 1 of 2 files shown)


eitsupi:
@dpprdan @benz0li Could you take a look at this?

eitsupi:
https://github.com/rocker-org/rocker-versioned2/runs/5991065015?check_suite_focus=true#step:5:6995

Apparently the locale is not set correctly...

```log
 #11 788.3 > sessionInfo()
#11 788.3 R version 4.0.0 (2020-04-24)
#11 788.3 Platform: x86_64-pc-linux-gnu (64-bit)
#11 788.3 Running under: Ubuntu 20.04.4 LTS
#11 788.3 
#11 788.3 Matrix products: default
#11 788.3 BLAS:   /usr/lib/x86_64-linux-gnu/blas/libblas.so.3.9.0
#11 788.3 LAPACK: /usr/lib/x86_64-linux-gnu/lapack/liblapack.so.3.9.0
#11 788.3 
#11 788.3 locale:
#11 788.3 [1] C
#11 788.3 
#11 788.3 attached base packages:
#11 788.3 [1] stats     graphics  grDevices utils     datasets  methods   base     
#11 788.3 
#11 788.3 loaded via a namespace (and not attached):
#11 788.3 [1] compiler_4.0.0
```

dpprdan:
LGTM!

(However, I see some errors in the [inspect image step](https://github.com/rocker-org/rocker-versioned2/runs/5991888058?check_suite_focus=true#step:6:8))

eitsupi:
Thank you both.

> (However, I see some errors in the [inspect image step](https://github.com/rocker-org/rocker-versioned2/runs/5991888058?check_suite_focus=true#step:6:8))

This is a Makefile issue (and one that has nothing to do with the target executed this time), so it is irrelevant to the image that is built.

PR #424, opened 2022-04-15
Change the rule for selecting Ubuntu version of base image

eitsupi:
Related to #282

Since RStudio and others may not be able to support the new LTS version of Ubuntu immediately after its release, a 90-day grace period will be set so that Ubuntu will not be set as the base image immediately after its release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,10 @@ Compared to `r-base`, this stack:
 
 - Builds on Ubuntu LTS rather than Debian and system libraries are tied to the Ubuntu version.
   Images will use the most recent LTS available at the time when the corresponding R version was released.
-  Thus all 4.0 images are based on Ubuntu 20.04 (`ubuntu:focal`).
+  - Since compatibility problems are likely to occur immediately after the release of a new Ubuntu LTS,
+    the version to be used is the one that is at least 90 days past release.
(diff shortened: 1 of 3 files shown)


eitsupi:
@cboettig @noamross I have set it at 90 days for now, but have no specific rationale. I would be glad to hear your opinion.
And, it would also be appreciated if you could proofread the document...

cboettig:
All the textual changes look good.  

I agree that having a 90-day window makes sense even if it's hard to justify the precise duration.  Among other things, after the 20.04 release NVIDIA took about that time before releasing 20.04-based cuda images which we use as our base in the ML stack.  not directly related to rocker, but I believe platforms like GitHub Actions saw a similar lag before providing 20.04-based VMs.  It will definitely help to have the additional window.  

Thanks for handling this!

cboettig (approved):

cboettig:
Also, just a note that another significant reason to for the wait is to give enough time for RSPM to provide `jammy` based binary builds.  I don't see any information on how long RStudio anticipates before adding this, but 90 days seems good for that too.

eitsupi:
Thank you for your review.
I will merge this and prepare for the R4.2.0 release.

PR #437, opened 2022-04-30
[CI] [Reports] Don't delete image lists to avoid erasing information of old groups

eitsupi:
The `bionicimages` group of 4.0.0 has been deleted by #436.
However, if we do not delete the image lists from the wiki, the information on the last built images can be referenced from the wiki.
(I meant to make this change in #436 and forgot)


eitsupi:
It worked fine on my fork, so I'm merging it.

PR #438, opened 2022-05-01
Allows `install_binder.sh` to be independently executable

eitsupi:
As mentioned in the following comment, currently `install_binder.sh` cannot be executed unless `install_python.sh` has been executed in advance.
This is different from the behavior of other installation scripts and should be changed so that it can be executed independently.
https://github.com/rocker-org/rocker-versioned2/pull/429#discussion_r855956332

### DIFF
--- a/scripts/install_binder.sh
+++ b/scripts/install_binder.sh
@@ -16,7 +16,9 @@ function apt_install() {
     fi
 }
 
-apt_install git sudo libzmq3-dev
+apt_install \
+    sudo \
+    libzmq3-dev
 
 # set up the default user if it does not exist
 if ! id -u "${NB_USER}" >/dev/null 2>&1; then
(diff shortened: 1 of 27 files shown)


scripts/install_binder.sh
```diff
@@ -1,31 +1,54 @@
...
+# https://github.com/rocker-org/rocker-versioned2/issues/428
 PATH=/opt/pyenv/bin:${PATH}
-
-# And set ENV for R! It doesn't read from the environment...
-echo "PATH=${PATH}" >>"${R_HOME}/etc/Renviron.site"
 echo "export PATH=${PATH}" >>"${WORKDIR}/.profile"
 
 cd "${WORKDIR}"
 ## This gets run as user
 sudo -u "${NB_USER}" python3 -m venv "${PYTHON_VENV_PATH}"
```
eitsupi, line 43: I do not understand the need for this line.
My understanding is that this environment is not used within this script...

eitsupi: It seems that `install_python.sh` also has unused code related to `venv` or `virtualenv`.
Do you think it is OK to delete these all together as well?

cboettig: probably. do you want to tackle updates to install_python.sh here?

eitsupi: This PR set up tests for each script and it make it easier future changes, so I would like to remove the virtual envs at another time.
I would like to merge this PR into #441 first to make sure it passes all tests and then into the default branch.

PR #439, opened 2022-05-01
Remove TERM=xterm from Dockerfiles' ENV

eitsupi:
Since it has already been set in the base images, there appears to be no need for setting in our Dockerfiles.

```shell
$ docker run --rm -it ubuntu bash
$ echo $TERM
xterm
```


cboettig:
yup, LGTM!

cboettig (approved):

PR #440, opened 2022-05-02
[CI] Allow changing the base image for rocker_scripts tests, and some improvements

eitsupi:
- Add new arg `base_image` to change the base image for a test build.
- Upload test build inspect files.
- `fail-fast: false` not to stop build by troubles.


eitsupi:
It seems to be working well, so I will merge it.

PR #435, opened 2022-04-29
Support installing RStudio Server for Ubuntu 22.04

eitsupi:
Related to #282

It appears that `ubuntu:jammy` needs to install a different build of RStudio, which not for `ubuntu:bionic` and `ubuntu:focal`.
https://github.com/rstudio/rstudio/issues/9854#issuecomment-1112568060

PR #441, opened 2022-05-03
Improve testing and refactoring of rocker scripts

### LINKED ISSUE #442: GPG error occured in `install_cuda-10.1.sh`
An error occurred when I tested the execution of the script in #441.
https://github.com/rocker-org/rocker-versioned2/actions/runs/2263365304

The GPG key update seems to be the cause.
https://forums.developer.nvidia.com/t/gpg-error-http-developer-download-nvidia-com-compute-cuda-repos-ubuntu1804-x86-64/212904
https://developer.nvidia.com/blog/updating-the-cuda-linux-gpg-repository-key/

eitsupi:
Add test cases and refactor each script.

- Add tests for the following scripts.
  With this change, all scripts used in pre-built images (except extra images) will be tested.
  - `install_cuda-10.1.sh`
  - `install_cuda-11.1.sh`
  - `config_R_cuda.sh`
 - Refactor each script
  (mainly removing unnecessary `apt update`, which can affect run time, especially in places far from the United States)

This PR includes the following PR tasks
(Originally, I think it is better to separate Pull Requests for each modification, but to avoid conflicts in the test definition file, I have put them together in one PR.)

- Update `install_rstudio.sh` for suporting `ubuntu:jammy` (#435)
- Update `install_binder.sh` to be able to run alone (#438)

And some other fixes.

- Add new GPG key for installing cuda packages. (fix #442)

### DIFF
--- a/.github/workflows/scripts-test.yml
+++ b/.github/workflows/scripts-test.yml
@@ -39,7 +39,8 @@ jobs:
       - name: Prepare build
         id: prep
         run: |
-          test_name="${{ matrix.base_image }}-${{ matrix.tag }}-${{ matrix.script_name }}-${{ matrix.script_arg }}"
+          lower_script_name=$(echo ${{ matrix.script_name }} | tr '[:upper:]' '[:lower:]')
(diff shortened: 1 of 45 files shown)


eitsupi:
It passed the tests and should work fine, so I will merge this.

PR #453, opened 2022-05-12
Feature/geospatial versions

cboettig:
proposed bump to geospatial versions


cboettig:
@eitsupi sorry I'm a bit confused about the lint checks on lines I didn't touch in this PR -- what am I missing?

eitsupi:
The results of the lint check are noted in the log. For example, the following line:
https://github.com/rocker-org/rocker-versioned2/runs/6416896580?check_suite_focus=true#step:4:82

```log
echo "options(repos = c(CRAN = '${CRAN}'))" >>${R_HOME}/etc/Rprofile.site
                                              ^-------^ SC2086 (info): Double quote to prevent globbing and word splitting.
```

Detailed rules for shellcheck can be found on the wiki of the repository.
https://github.com/koalaman/shellcheck/wiki/SC2086

Is 8c3dad97e22839b5893a2045c1c3f722442d80ca the only content you want to edit in this PR?
If so, can I undo the other changes and force push?
If we don't edit any other files, I think linter will have little to complain about.

PR #455, opened 2022-05-20
Refactoring install scripts for installing geospatial packages

### LINKED ISSUE #376: Update versions of OSGEO libs on rocker/geospatial
Currently `rocker/geospatial` has the following OSGEO system libraries:

```bash
docker run --rm -it rocker/geospatial /bin/bash
```

Currently that results in this:

```
sf::sf_extSoftVersion()
          GEOS           GDAL         proj.4 GDAL_with_GEOS     USE_PROJ_H 
       "3.8.0"        "3.0.4"        "6.3.1"         "true"         "true" 
          PROJ 
       "6.3.1" 
```

PROJ in particular is out of date now and some packages (notably `cartopy`) depend on 8.0.0 or greater. I will try the OSGeo-dev container, that may solve the specific issue I'm hitting.

eitsupi:
- [x] Remove unused scripts
- [x] Run `install_geospatial.sh` in `install_geospatial_unstable.sh`
- [x] Get the latest versions of PROJ, GDAL, GEOS automatically (close #376)
- [x] Rename `dev_osgeo.sh` to `install_dev_osgeo.sh`
- [x] Other minor updates


eitsupi:
@cboettig Do you think it is acceptable to delete the following scripts, including the one you were going to delete in #453, as in 184410c?

- add_ubuntugis.sh
- bh-gdal.sh
- bh-proj-gdal_only.sh
- bh-proj.sh
- install_gdal_source.sh
- install_proj.sh

eitsupi:
@cboettig Thank you for your confirmation. I have two other interests.

1. Since `install_geospatial_unstable.sh` appears to be a script that assumes that `install_geospatial.sh` will be executed immediately after, it would be better to call `install_geospatial.sh` within `install_geospatial_unstable.sh`.
2. The name `dev_osgeo.sh` appears to deviate from the naming scheme of the other scripts. Perhaps something like `install_dev_osgeo.sh` would be better.

mdsumner:
ah cool, clears up a lot for me 🤗🙏

scripts/dev_osgeo.sh
```diff
@@ -1,8 +1,8 @@
 #!/bin/bash
 set -e
 
-CRAN=${CRAN_SOURCE:-https://cran.r-project.org}
-echo "options(repos = c(CRAN = '${CRAN}'), download.file.method = 'libcurl')" >>"${R_HOME}/etc/Rprofile.site"
+CRAN=${CRAN_SOURCE:-"https://cloud.r-project.org"}
+echo "options(repos = c(CRAN = '${CRAN}'))" >>"${R_HOME}/etc/Rprofile.site"
```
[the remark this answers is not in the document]
eitsupi: Is it necessary to source install R packages unrelated to proj, gdal, and geos?
I think we can save build time by doing a binary install of as many R packages as possible.
Do you know specifically which R packages only need to be source installed?

scripts/install_dev_osgeo.sh
```diff
@@ -0,0 +1,166 @@
...
+git clone --depth 10 https://github.com/r-spatial/sf.git
+git clone --depth 10 https://github.com/r-spatial/lwgeom.git
+git clone --depth 10 https://github.com/r-spatial/stars.git
+
+R CMD build --no-build-vignettes --no-manual lwgeom
+R CMD build --no-build-vignettes --no-manual sf
+R CMD INSTALL sf
+R CMD INSTALL lwgeom
+R CMD build --no-build-vignettes --no-manual stars
+R CMD INSTALL stars
```
eitsupi, line 162: @cboettig Is there any reason not to remove these downloaded repositories after package installation?

cboettig: nope, that was an oversight! better to remove I think.

scripts/dev_osgeo.sh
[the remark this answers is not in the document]
cboettig: yeah good question, I was wondering about that as well.  Of course we could one-off install these from source at docker build time without changing the default CRAN repo away from RSPM at all, which would be fine I think.  The difficulty is only if users later upgrade any package binding any of these 3 libraries (which includes some packages not yet on our install list in this script, e.g. `terra`, the successor to `raster`, and will no doubt include other packages in the future).  

I don't know of a good way to ensure R will only use RSPM binaries from packages where the libs come from the official ubuntu repos but source install those packages that have libs built from source (or ppas).  Maybe @jeroen has some advice?  Otherwise I think just turning off RSPM here is safest.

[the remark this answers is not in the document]
eitsupi: Changing the default repository with this script would make sense as you describe.
My concern is that the script seems to install too many R packages (no dependencies with proj, gdal, geos).
Do you think these can be removed from this script?
Or is it possible to install these packages before changing the configuration to do source install?
(I assume most of the packages are already installed in the pre-built image since it runs on `rocker/verse:latest`).

https://github.com/rocker-org/rocker-versioned2/blob/25c29ffb2fc14f21f50a0af02c132ba7bf06015c/scripts/install_dev_osgeo.sh#L125-L151

scripts/install_dev_osgeo.sh
eitsupi: Another thing I was wondering: is there a reason you are using options `--no-build-vignettes --no-manual` ?

scripts/install_dev_osgeo.sh
```diff
@@ -0,0 +1,177 @@
...
+wget http://download.osgeo.org/geos/geos-"${GEOS_VERSION}".tar.bz2
+bzip2 -d geos-*bz2
+tar xf geos*tar
+rm geos*tar
+cd geos*
+./configure
+make
+make install
+cd ..
+ldconfig
```
eitsupi, line 122: The following comment mentions the relationship between GDAL and GEOS, is this the correct order of build?
https://github.com/rocker-org/rocker-versioned2/issues/376#issuecomment-1059954786

build/make-stacks.R
```diff
@@ -121,6 +121,17 @@ library(gert)
...
+  gert::git_remote_ls(remote = remote_repo) |>
+    dplyr::pull(ref) |>
+    stringr::str_subset(r"(^refs/tags/v?(\d+\.)*\d+$)") |>
+    stringr::str_extract(r"((\d+\.)*\d+$)") |>
+    package_version() |>
+    sort() |>
+    utils::tail(1) |>
+    as.character()
+}
+
```
cboettig, line 134: very nice! thanks for adding this :smile:

dockerfiles/geospatial-dev-osgeo_4.2.0.Dockerfile
```diff
@@ -6,7 +6,7 @@ LABEL org.opencontainers.image.licenses="GPL-2.0-or-later" \
       org.opencontainers.image.authors="Carl Boettiger <cboettig@ropensci.org>"
 
 ENV PROJ_VERSION=9.0.0
-ENV GDAL_VERSION=3.4.1
+ENV GDAL_VERSION=3.5.0
 ENV GEOS_VERSION=3.10.2
 
-RUN /rocker_scripts/dev_osgeo.sh
+RUN /rocker_scripts/install_dev_osgeo.sh
```
cboettig, line 12: :+1:

scripts/install_dev_osgeo.sh
cboettig: thanks for catching that.  We should build GEOS first.

scripts/install_dev_osgeo.sh
```diff
@@ -0,0 +1,177 @@
...
+    protolite \
+    tmap \
+    mapview \
+    odbc \
+    pool \
+    rmarkdown \
+    RPostgres \
+    spatstat \
+    stars
+
```
cboettig, line 152: this list is just copied from install_geospatial.sh package list, and may not be in sync with it.  Some of these are slated for deprecation (rgdal, rgeos) or succeeded by other packages, and a good number don't bind the geospatial C libs so could be installed from binary (though some of those will pull in packages that do, so might need testing.  

I don't think we need to change this now, but just something for us to think about going forward -- could get some input from the spatial devs and revise this list with some of the newer state-of-the-art at the same time.

scripts/install_dev_osgeo.sh
cboettig: Ah, that was probably a copy-over from CI script used in stars; we should probably build the vignettes and manual.  Good catch.

cboettig (approved):

scripts/install_dev_osgeo.sh
eitsupi: @cboettig 

At this time, the R packages installed by the two scripts appear to be quite different. The following packages are only installed by dev_osgeo.

```
covr, DBI, dplyr, knitr, maps, microbenchmark, odbc, pool, protolite, rmarkdown, RPostgres, RPostgreSQL, RSQLite, testthat, tibble, tidyr, units
```

The following packages are installed only in geospatial.

```
BiocManager, classInt, deldir, geoR, gstat, hdf5r, lidR, mapdata, ncdf4, proj4, RandomFields, RColorBrewer, rlas, RNetCDF, spacetime, spatialreg, spdep, terra, tidync
```

I would suggest removing packages that exist only in dev_osgeo.
Most of these seem to be essentially irrelevant to dev_osgeo.

scripts/install_dev_osgeo.sh
eitsupi: I too was concerned that the build method was different from that described on the official website, but I hesitated to change the build method because this script has too long an execution time to be build tested.

Just a question: why am I still able to build with a different build method than the official site?

eitsupi: @rsbivand Thanks for pointing that.
I have created a new issue #458 because it probably cannot be fixed within this PR.

eitsupi:
I'll merge this in for now, as there may still be improvements to be made.......

scripts/experimental/install_dev_osgeo.sh
```diff
@@ -0,0 +1,177 @@
...
+
+function url_latest_gh_released_asset() {
+    wget -qO- "https://api.github.com/repos/$1/releases/latest" | grep -oP "(?<=\"browser_download_url\":\s\")https.*\.tar.gz(?=\")"
+}
+
+export DEBIAN_FRONTEND=noninteractive
+apt_install \
+    gdb \
+    git \
+    libcairo2-dev \
```
cboettig, line 36: @eitsupi can you add in cmake here? Optional for the current release but will be needed for gdal going forward. Thanks @rsbivand .

cboettig:
@eitsupi maybe we do want tot keep the no-build-vignettes option for now, seems that `stars` vignettes aren't building? https://github.com/rocker-org/rocker-versioned2/runs/6560429916?check_suite_focus=true#step:5:10081

mdsumner:
excellent, lovely to see this detail discussed so openly! I'm probably unable to try this out again until next week but I'll be using this a lot, thanks 😍

scripts/experimental/install_dev_osgeo.sh
eitsupi: I think cmake is (was) already included. look at line 64.

edzer:
> stars vignettes aren't building

it seems the packages that stars `Suggests:` aren't installed, so it makes sense to not try to build these vignettes.

eitsupi:
I've opened a new PR #459.

PR #456, opened 2022-05-20
[CI] Push images to GitHub Container registry too

### LINKED ISSUE #371: dockerhub rate limiting
We are currently using quay.io to handle docker containers built off of rocker/r-ver:3.6.0 (our container requires the libgfortran3 library or we would have updated to >4.0).  We're running into an issue with the docker hub pull rate limits that started at the end of November. I see a (defunct?) repository on quay.io called rocker (https://quay.io/repository/rocker/r-ver). Is this associated with the group that maintains this GitHub repository? If so, it would be a huge help to us to get a successful build that we could use on quay.io and circumvent the rate limits. Or should we find a different solution since the quay.io repo won't be updated? Thanks!

eitsupi:
Close  #371

Add tags like `ghcr.io/rocker-org/r-ver:latest` to push the same images to GitHub Container registry as to DockerHub.
Of course, the GitHub Actions workflow files also need a little updating to push images to ghcr.

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -50,10 +50,17 @@ jobs:
       matrix: ${{fromJson(needs.generate_matrix.outputs.matrix)}}
     steps:
       - uses: actions/checkout@v3
-      - uses: docker/login-action@v2
+      - name: Login to Docker Hub
+        uses: docker/login-action@v2
         with:
           username: ${{ secrets.DOCKER_USER }}
           password: ${{ secrets.DOCKER_PASSWORD }}
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v2
+        with:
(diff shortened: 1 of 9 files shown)


cboettig (approved):

cboettig:
Looks reasonable to me. We might want to push the older tags as well for completeness but probably not a priority. Does this need any further config on my end for the GH registry?

eitsupi:
In the end, no additional steps were needed.
I simply added a step to log into GHCR on the workflow and added the tags and it worked.
(I did change the settings to make images registered in this repository invisible a long time ago because it was confusing, but that has nothing to do with the tags that now pushes).

eitsupi:
We can verify that images with the exact same digest are available on DockerHub and GitHub.

image
image

PR #457, opened 2022-05-20
Patch/gdal

cboettig:


eitsupi:
The Dockerfile must also be updated.

Could you also confirm the contents of #455?

cboettig:
Thanks - isn't the Dockerfile updated automatically by the script though?  Or should I run `./build/make-stacks.R` locally?

Workflow-wise, would it be easier to fold these changes into #455 ?

eitsupi:
Yeah, unfortunately for now the Dockerfile won't update unless you do `make setup` locally...

If you are willing, I would like to take the content of this PR and revise it together with #455.
I am going to bed today but will try to finish #455 tomorrow so that we can merge and build.

cboettig:
@eitsupi sounds like a plan, thank you! I'll close this and thanks for folding it into #455

PR #459, opened 2022-05-24
Fix dev-osgeo installation

eitsupi:
Improve build of  `rocker/geospatial:dev-osgeo`


cboettig (changes requested):
@eitsupi can you just add `cmake` to the apt-get install dependency list in dev-osgeo?

eitsupi:
It is already exist.

PR #461, opened 2022-05-26
[Docs] Enhanced explanation of special tags

eitsupi:
I thought explanations like https://github.com/rocker-org/rocker-versioned2/issues/376#issuecomment-1132072742 should be in the README.

We also saw a few questions about `ubuntu:jammy`, so I also specified that the images with the `devel` tag are based on `ubuntu:latest`.


eitsupi:
Merge this for now. If there are any problems, please let me know.

PR #467, opened 2022-05-29
Add strip step to reduce container image size

### LINKED ISSUE #340: Binary builds from RSPM blow up image size
Reason: The dynamic libraries are not stripped.

`rocker/tidyverse` (Ubuntu 20.04, RSPM, binary install):

```bash
root@89f0a27ec269:/usr/local/lib/R/site-library/vroom/libs# du -h vroom.so 
24M	vroom.so
root@89f0a27ec269:/usr/local/lib/R/site-library/vroom/libs# file vroom.so
vroom.so: ELF 64-bit LSB shared object, x86-64, version 1 (GNU/Linux), dynamically linked, BuildID[sha1]=88a2384029c2331d1d8f91274db23fb30edb9668, with debug_info, not stripped
```

`custom build` (Debian 11, CRAN, source install):

```bash
root@bfc0aad5ae64:/usr/local/lib/R/site-library/vroom/libs# du -h vroom.so 
848K	vroom.so
root@bfc0aad5ae64:/usr/local/lib/R/site-library/vroom/libs# file vroom.so
vroom.so: ELF 64-bit LSB shared object, x86-64, version 1 (GNU/Linux), dynamically linked, BuildID[sha1]=f6d17acb2cf4f8e75af5e97dcb7a17297ff5f9ee, stripped
```

Total size of `/usr/local/lib/R/site-library`:

*  `rocker/tidyverse`: 732 MB
*  `custom build`: 198 MB

@eddelbuettel Can you think of a reason why the dynamic libraries of the RSPM binaries are not stripped?

eitsupi:
(Maybe) close #340

### DIFF
--- a/scripts/install_geospatial.sh
+++ b/scripts/install_geospatial.sh
@@ -84,3 +84,7 @@ R -e "BiocManager::install('rhdf5')"
 # Clean up
 rm -rf /var/lib/apt/lists/*
 rm -r /tmp/downloaded_packages
+
+## Strip binary installed lybraries from RSPM
+## https://github.com/rocker-org/rocker-versioned2/issues/340
+strip /usr/local/lib/R/site-library/*/libs/*.so
--- a/scripts/install_julia.sh
+++ b/scripts/install_julia.sh
@@ -56,3 +56,7 @@ julia --version
 # Clean up
 rm -rf /var/lib/apt/lists/*
 rm -rf /tmp/downloaded_packages
+
+## Strip binary installed lybraries from RSPM
+## https://github.com/rocker-org/rocker-versioned2/issues/340
+strip /usr/local/lib/R/site-library/*/libs/*.so
--- a/scripts/install_python.sh
+++ b/scripts/install_python.sh
@@ -75,3 +75,7 @@ chown :staff "${PYTHON_VENV_PATH}"
 # Clean up
 rm -rf /var/lib/apt/lists/*
 rm -rf /tmp/downloaded_packages
+
+## Strip binary installed lybraries from RSPM
+## https://github.com/rocker-org/rocker-versioned2/issues/340
+strip /usr/local/lib/R/site-library/*/libs/*.so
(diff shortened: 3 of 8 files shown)


eitsupi:
Focusing on the image size included in the test results, it seems that the size of the image running `install_tidyverse.sh` on `rocker/r-ver:latest` has decreased from 1.7GB to 1.2GB🎉

@cboettig @eddelbuettel @benz0li
Does this change look good? Sorry I have no knowledge of strip at all...

benz0li:
> [...] the size of the image running `install_tidyverse.sh` on `rocker/r-ver:latest` has decreased from 1.7GB to 1.2GB🎉

This size reduction seems about right.

> [...] Does this change look good? Sorry I have no knowledge of strip at all...

@eitsupi I'm no expert either... but you could reduce the image sizes even further by using `make install-strip` at https://github.com/rocker-org/rocker-versioned2/blob/master/scripts/install_R_source.sh#L136 for the R installation itself.

eitsupi:
Thanks for the reply and suggestion!

According to the manual, it could be reduced by about 20MB?
This is certainly worth considering, but the effect is small compared to the R package's strip and could be corrected with another PR.

> On ‘x86_64’ Linux a typical reduction in overall size was from 92MB to 66MB.

https://cran.r-project.org/doc/manuals/r-release/R-admin.html#Debugging-Symbols

benz0li:
> According to the manual, it could be reduced by about 30MB? This is certainly worth considering, but the effect is small compared to the R package's strip [...].

Agreed. In absolute terms, this is a minor gain compared to stripping the R packages' `libs`. In relative terms, this is a 28% size reduction for the R installation itself.

cboettig:
LGTM :tada:

eitsupi:
It appears that the size of the images could be reduced. Thanks @benz0li !

before https://github.com/rocker-org/rocker-versioned2/wiki/geospatial_8992e1753bff
after https://github.com/rocker-org/rocker-versioned2/wiki/geospatial_8bbd4132651f

PR #482, opened 2022-06-08
Make texlive's path to be unrelated to CPU architectures

### LINKED ISSUE #478: Hard-coded arch for texlive in PATH
https://github.com/rocker-org/rocker-versioned2/blob/76c47671f25c76a0c99c9ac31c202ab8afde3264/dockerfiles/verse_4.2.0.Dockerfile#L9

benz0li:
- Remove (hard-coded) ARCH in PATH

Fixes #478

### DIFF
--- a/dockerfiles/verse_4.2.0.Dockerfile
+++ b/dockerfiles/verse_4.2.0.Dockerfile
@@ -6,7 +6,7 @@ LABEL org.opencontainers.image.licenses="GPL-2.0-or-later" \
       org.opencontainers.image.authors="Carl Boettiger <cboettig@ropensci.org>"
 
 ENV CTAN_REPO=https://mirror.ctan.org/systems/texlive/tlnet
-ENV PATH=$PATH:/usr/local/texlive/bin/x86_64-linux
+ENV PATH=$PATH:/usr/local/texlive/bin/linux
 ENV QUARTO_VERSION=latest
 
 RUN /rocker_scripts/install_verse.sh
--- a/dockerfiles/verse_4.2.0.Dockerfile
+++ b/dockerfiles/verse_4.2.0.Dockerfile
@@ -6,7 +6,7 @@ LABEL org.opencontainers.image.licenses="GPL-2.0-or-later" \
       org.opencontainers.image.authors="Carl Boettiger <cboettig@ropensci.org>"
 
 ENV CTAN_REPO=https://mirror.ctan.org/systems/texlive/tlnet
-ENV PATH=$PATH:/usr/local/texlive/bin/linux
+ENV PATH=$PATH:/usr/local/texlive/bin/x86_64-linux
 ENV QUARTO_VERSION=latest
 
 RUN /rocker_scripts/install_verse.sh
--- a/scripts/install_texlive.sh
+++ b/scripts/install_texlive.sh
@@ -43,7 +43,7 @@ cd ..
 rm -rf install-tl-*
 rm /tmp/texlive-profile.txt
 
-ln -rs /usr/local/texlive/bin/${ARCH}-linux /usr/local/texlive/bin/linux
+ln -rs /usr/local/texlive/bin/"${ARCH}"-linux /usr/local/texlive/bin/linux
 
 tlmgr update --self
 tlmgr install latex-bin luatex xetex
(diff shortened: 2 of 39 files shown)


benz0li:
@eitsupi The symbolic link is created before any `tlmgr` action is executed. This is required, because installing `context` runs `mtxrun --generate` post-installation.  
ℹ️ Without the symbolic link, `mtxrun` is not found / not in `PATH`.

benz0li:
https://github.com/rocker-org/rocker-versioned2/blob/ae741564c8c965672d329a94e7930c4e4ba98809/scripts/install_texlive.sh#L31

might not be required, because `ENV` is set beforehand in the Dockerfile.

dockerfiles/verse_4.2.0.Dockerfile
```diff
@@ -6,7 +6,7 @@ LABEL org.opencontainers.image.licenses="GPL-2.0-or-later" \
       org.opencontainers.image.authors="Carl Boettiger <cboettig@ropensci.org>"
 
 ENV CTAN_REPO=https://mirror.ctan.org/systems/texlive/tlnet
-ENV PATH=$PATH:/usr/local/texlive/bin/x86_64-linux
+ENV PATH=$PATH:/usr/local/texlive/bin/linux
```
eitsupi, line 9: Do not edit this Dockerfile by hand.
Please check the document. https://github.com/rocker-org/rocker-versioned2/blob/ae741564c8c965672d329a94e7930c4e4ba98809/build/README.md

eitsupi (changes requested):
Thank you for taking a look at this.
However, there may be no need to update these Dockerfiles at this stage as these can only be built on linux/arm64.

I also wonder if adding the symbolic link to a location already added to the PATH would eliminate the need to update the PATH.

dockerfiles/verse_4.2.0.Dockerfile
benz0li: > However, there may be no need to update these Dockerfiles at this stage as these can only be built on linux/arm64.

You meant linux/amd64, correct? Just looking ahead for future linux/arm64/v8 builds

> I also wonder if adding the symbolic link to a location already added to the PATH would eliminate the need to update the PATH.

No. The link – not the target – needs to be in PATH.

> Please check the document. https://github.com/rocker-org/rocker-versioned2/blob/ae741564c8c965672d329a94e7930c4e4ba98809/build/README.md

I will check and update the pull request accordingly.

scripts/install_texlive.sh
```diff
@@ -45,6 +43,8 @@ cd ..
 rm -rf install-tl-*
 rm /tmp/texlive-profile.txt
 
+ln -rs /usr/local/texlive/bin/${ARCH}-linux /usr/local/texlive/bin/linux
```
eitsupi, line 46: Is there a reason you are specifying the `-r` option?

benz0li: In such a case I prefer a relative link.

eitsupi (changes requested):
Unfortunately, the build test failed.
And, suggestion to avoid linter complaining.

scripts/install_texlive.sh
eitsupi: Note that other scripts does not use the `-r` option now.

https://github.com/rocker-org/rocker-versioned2/blob/ae741564c8c965672d329a94e7930c4e4ba98809/scripts/install_pandoc.sh#L53

benz0li: > In such a case I prefer a relative link.

such a case = a symbolic link within the same directory.

eitsupi: I see. I have no preference myself.
Certainly the `-r` option might be reasonable for the same directory.

Given the possibility of repeated execution, add the `-f` option seems a good idea.

scripts/install_texlive.sh
```diff
@@ -28,8 +28,6 @@ option_doc 0
 option_src 0
 EOF
 
-export PATH="${PATH}:/usr/local/texlive/bin/${ARCH}-linux/"
```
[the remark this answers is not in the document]
benz0li: Isn't `/usr/local/texlive/bin/linux` added to PATH via ENV from the JSON-files?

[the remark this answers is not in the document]
eitsupi: As described in the README, rocker script execution does not assume PATH updating.
https://github.com/rocker-org/rocker-versioned2#rocker-scripts

benz0li:
@eitsupi Thanks for reviewing.

Issue #476, opened 2022-06-01
Adding Packages in Docker Compose within Portainer Stack


frabau1:
Is it possible to add packages via a Docker Compose statement in a Portainer stack?  I typically deploy my docker containers as  stacks and I’ve written the following code to deploy this stack:

----

version: '3'

services:
  shiny:
    image: rocker/shiny-verse:latest
    environment: 
      #RUN: "install.packages('sf')"
      RUN: "install2.R sf"
    restart: always
    networks: 
       - shinyservers
    ports:
       - "3838:3838"
       - "8787:8787"
    volumes:
       - "/volume1/docker/Shiny/Logs:/var/log/shiny-server"
       - "/volume1/docker/Shiny/Apps:/srv/shiny-server"
networks:
   shinyservers:
     driver: bridge 

----

My issue revolves around the environmental variables with RUN commands above.   I’m not familiar with the syntax to install additional packages when deploying Rocker as a stack, and have been unable to find any examples.  The documentation on the Rocker website references command line code https://www.rocker-project.org/use/extending/, which is what I attempted to insert above in a failed experiment.

I have successfully deployed test programs that do not require any additional packages using the rest of the code  above, however the project I’m working on needs the ’sf’ package in addition to those bundled with the shiny-verse image.  I’ve experimented with additional syntax beyond what I inserted above without success.  

Is there a way to deploy a Rocker stack and include additional packages?  Any guidance would be greatly appreciated.

cboettig:
I'm not familiar with `Portainer`, but I think the standard approach to extending Rocker images would be to write a Dockerfile that is derived from the desired Rocker image with the desired additional RUN commands, and then use the `build` directive in your docker compose yml.

eitsupi:
Creating the following `compose.yaml` and `Dockerfile` should work (not tested).

```yml
services:
  shiny:
    build: .
    ports:
      - 3838:3838
```

```dockerfile
FROM rocker/shiny-verse:latest
RUN install2.r --error -n -1 sf
```

Please check https://docs.docker.com/compose/gettingstarted/.

(Note that we are currently considering adding examples of such usage to the Rocker Project site)

frabau1:
I have successfully built a package using the docker file code provided by @eitsupi , however I am getting an error in my log stating that the package isn’t available when I run my shiny app:

── Attaching packages ─────────────────────────────────────── tidyverse 1.3.1 ──
[...]

PR #483, opened 2022-06-10
Add test for `install_shiny_server.sh` on `rocker/r-bspm:20.04`

eitsupi:
Related to #476

Users may find it easier to install Shiny Server on `rocker/r-bspm` than to install the R packages in `rocker/shiny-verse`.

We can expect the script to install Shiny Server to work with `rocker/r-bspm:20.04` and so on, so I would like to add a test to confirm that it works.
(Of course, in the future we will need to document which scripts have been tested to work with `rocker/r-bspm:20.04` etc.)


eitsupi:
It seems that `apt-get` update was skipped due to remaining cache and the installation test failed.
I will take a look the upstream Dockerfile later.

https://github.com/rocker-org/rocker-versioned2/blob/77e4b1c905898e6fb88ce841a9303a7cfce3c6df/scripts/install_shiny_server.sh#L12-L14

PR #511, opened 2022-07-23
Remove devscripts expect the checkbashisms script, and remove lsb-release from `rocker/r-ver`

### LINKED ISSUE #510: Remove `devscripts` installation from install R scripts to reduce container size
Tracking this discussion, I see that `rocker/drd` doesn't like the devscripts dependencies (like python3) and only extracts `checkbashisms` and includes it in the source code.

https://github.com/rocker-org/rocker-versioned/pull/167#issuecomment-549041154
https://github.com/rocker-org/drd/blob/06ab465832ab7c77335d16f9e934d03bdb0493c4/checkbashisms

I suppose we could do the same thing here.

_Originally posted by @eitsupi in https://github.com/rocker-org/rocker-versioned2/issues/507#issuecomment-1190315281_

eitsupi:
Close #510

**ToDo**

- [x] Update the GitHub Actions workflow for testing `install_R_source.sh` to test rocker scripts on new `rocker/r-ver`. (#523)
- [x] Update NEWS.md

### DIFF
--- a/.github/workflows/r-build-test.yml
+++ b/.github/workflows/r-build-test.yml
@@ -55,7 +55,7 @@ jobs:
     needs: build
     runs-on: ubuntu-latest
     strategy:
-      fail-fast: true
+      fail-fast: false
       matrix:
         tag:
           - "4.0.0"
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,12 @@
 # News
 
+## 2022-11
+
+### Changes in pre-built images
+
+- `lsb-release` and `python3` (a dependency of `lsb-release`), are no longer included in `rocker/r-ver`.
+  ([#511](https://github.com/rocker-org/rocker-versioned2/pull/511))
+
 ## 2022-10
 
 ### Changes in pre-built images
--- a/scripts/install_R_source.sh
+++ b/scripts/install_R_source.sh
@@ -34,7 +34,6 @@ fi
 apt-get install -y --no-install-recommends \
     bash-completion \
     ca-certificates \
-    devscripts \
     file \
     fonts-texgyre \
     g++ \
@@ -60,6 +59,7 @@ apt-get install -y --no-install-recommends \
 
 BUILDDEPS="curl \
     default-jdk \
+    devscripts \
     libbz2-dev \
     libcairo2-dev \
     libcurl4-openssl-dev \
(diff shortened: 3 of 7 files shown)


eitsupi:
Unfortunately, the image built from 4a1df66ac29940f68c19b49a59b674e86bbebbc4 does not appear to have reduced image size. 🤔

eitsupi:
Perhaps we need to rethink the minimum package requirements: python3 will be installed even if devscripts is not installed (listed below).

https://github.com/rocker-org/rocker-versioned2/blob/4a1df66ac29940f68c19b49a59b674e86bbebbc4/scripts/install_R_source.sh#L34-L57

cboettig:
yeah, I'm still relatively skeptical that there are major savings to be had here, especially that will persist after a user also adds commonly used packages etc, at least while remaining based on Ubuntu.  @gaborcsardi's Alpine based https://github.com/r-hub/r-minimal is probably a better option for environments where saving 10s of MB is critical.

benz0li:
@eitsupi The problematic line is

https://github.com/rocker-org/rocker-versioned2/blob/4a1df66ac29940f68c19b49a59b674e86bbebbc4/scripts/install_R_source.sh#L16

ℹ️ `lsb-release` depends on `python3`.

benz0li:
> yeah, I'm still relatively skeptical that there are major savings to be had here

@cboettig It's not only about reducing image size, but also reducing the number of [unneeded] packages.

Installing `devscripts` on bare `ubuntu:focal` results in 104 newly installed packages.

benz0li:
You could use `UBUNTU_VERSION=$(. /etc/os-release && echo $UBUNTU_CODENAME)` at

https://github.com/rocker-org/rocker-versioned2/blob/4a1df66ac29940f68c19b49a59b674e86bbebbc4/scripts/install_R_source.sh#L23

instead of [installing and] using `lsb_release`.

cboettig:
Thanks @benz0li , :100: with you on trimming unnecessary dependencies. The `lsb_release` is a good example. 

Having unneeded packages increases image size and maybe makes a bigger surface area of dependencies that could break later, as well as potential security issues.  Though for the latter two issues I think relying on the official ubuntu LTS repo sources is probably a better way to avoid either security or dependency issues than us hand-rolling dependencies.  I like both of the above work-arounds because it looks like they will still be based on the official repo sources, where we benefit from the dependency management and security patching of the Debian and Canonical maintainers.

eitsupi:
I've been writing shell scripts as Dev Container Features for these days and I've learned how good the method of loading `os-release` scripts is.
https://github.com/rocker-org/devcontainer-features/blob/cbee3bfdce891012e88ae87c0da2c925a6a11301/src/r-apt/install.sh#L34
https://github.com/rocker-org/devcontainer-features/blob/cbee3bfdce891012e88ae87c0da2c925a6a11301/src/r-apt/install.sh#L123-L129
Definitely a good thing to bring this here.

eitsupi:
The size of rocker/r-ver:devel was reduced from 836MB (bfe2d71) -> 808MB (c5a1e36) by deleting lsb-release.
However, lsb-release is a dependency of RStudio Server, so it seems that rocker/rstudio needs to be lsb-release installed.

And, Python 3 continues to be installed. I do not know where this came from.

PR #518, opened 2022-08-18
[Docs][CI] Change the link syntax on the wiki home from MediaWiki's one to Markdown's to reduce redering time on GitHub

### LINKED ISSUE #472: The wiki home took too long to reder
I think it is due to the long list of tags contained in the table, and it seems that the additional tags for GHCR are taking more longer to render as a result.

image

eitsupi:
close #472

Two syntaxes are available for linking on the GitHub wiki, but I have found that the MediaWiki syntax takes a long time to render and sometimes fails to render on wikis with large numbers of pages.

> - If your pages are rendered with Markdown, the link syntax is `[Link Text](full-URL-of-wiki-page)`.
> - With MediaWiki syntax, the link syntax is `[[nameofwikipage|Link Text]]`.

https://docs.github.com/en/communities/documenting-your-project-with-wikis/editing-wiki-content

So change the link syntax and update the wiki home so that it renders correctly.

### DIFF
--- a/build/reports/wiki_home.Rmd
+++ b/build/reports/wiki_home.Rmd
@@ -134,7 +134,7 @@ df_all |>
     R = version,
     ImageName = image_title,
     RepoTags = tags,
-    ID = stringr::str_c("[[", id, "|", report_name, "]]"),
+    ID = stringr::str_c("[", id, "](./", report_name, ")"),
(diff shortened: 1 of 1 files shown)


eitsupi:
It seems working fine on my fork, so I merge it.

Issue #520, opened 2022-08-23
improved devcontainer support?


cboettig:
Looking at the amazing work @eitsupi and @revodavid have done with devcontainer standard (particularly for deploy on codespaces, which looks rather promising in some contexts), I'm wondering if we should support more of that boilerplate here.  Not sure if that would b best as an additional image in the stack or (probably better) an additional bash script in `rocker_scripts` that can bundle up at least some of the additional things that get added to the [devcontainer Dockerfile](https://github.com/revodavid/devcontainers-rstudio/blob/main/.devcontainer/Dockerfile) and [debian script](https://github.com/revodavid/devcontainers-rstudio/blob/main/.devcontainer/library-scripts/common-debian.sh).  

(There's also a lot in there I haven't quite wrapped my head around or am unsure about.  e.g. see https://github.com/revodavid/devcontainers-rstudio/issues/1, I'm not sure that we want to use separate users `vscode` and `rstudio` on the same VM? seems to create a minor issue for git too?)  Thoughts / ideas / objections welcome!

eitsupi:
First of all, I think it is necessary for you to understand the status of devcontainer (microsoft/vscode-dev-containers#1589).

For a long time, devcontainer could only be handled from the proprietary vscode extension available only on Visual Studio Code, but Microsoft has recently moved this to an open source implementation.
https://github.com/devcontainers/spec
https://github.com/devcontainers/cli

In parallel, they are working on implementing a feature called devcontainer features, which will add functionality on the containers.
https://github.com/devcontainers/features

This is effectively a more generalized version of the Rocker Scripts that exist in this repository.
In other words, I believe that almost all of the contents of this repository could be migrated to the devcontainer features.
(devcontainer features spec still a work in progress and subject to destructive changes / I was going to try this in the near future in my personal repository)
If such a thing could be done, it would more generally accomplish what I was trying to do in #483.

For example, the contents of [the script](https://github.com/devcontainers/features/blob/f82307ad60514d1e912a9aa2b64204721fe955eb/src/common-utils/install.sh) can be imported into any debian-based container by building devcontainer from the following json file.

```json
"name": "my-project-devcontainer",
"image": "mcr.microsoft.com/devcontainers/base:ubuntu",  // Any generic, debian-based image.
"features": {
    "ghcr.io/devcontainers/features/common-utils:1": {
        "version": "latest"
    }
}
```

[...]

PR #534, opened 2022-09-07
Move RStudio user settings from `.rstudio/monitored/user-settings/user-settings` to `.config/rstudio/rstudio-prefs.json`

### LINKED ISSUE #63: RStudio preferences
With the RStudio 1.3.x being released and making use of the `rstudio-prefs.json` file, it may be necessary to adjust how the rstudio script handles reproducibility.

I've recently built a container (taking the lead from your amazing work) and experienced some issues when making use of the new `json` file setup for preferences.

The issue comes from this piece of code in the `install-studio.sh` script 

```
## Rocker's default RStudio settings, for better reproducibility
mkdir -p /home/rstudio/.rstudio/monitored/user-settings \
  && echo 'alwaysSaveHistory="0" \
          \nloadRData="0" \
          \nsaveAction="0"' \
          > /home/rstudio/.rstudio/monitored/user-settings/user-settings \
  && chown -R rstudio:rstudio /home/rstudio/.rstudio
```
It seems these settings cause conflict when a `json` settings file is present. I was experiencing the case that the settings that would load in the container would be set to save history, set to ask about saving RData and set to load RData.

When I removed the above mentioned code in favour of just the `json` file, the preferences I'd set in the file worked.

eitsupi:
Fix #63 and related to #520

RStudio Server 1.3 or later, the configuration file is `~/.config/rstudio/rstudio-prefs.json` instead of `~/.rstudio/monitored/user-settings/user-settings`.
`~/.rstudio/monitored/user-settings/user-settings` will be deleted when RStudio is started and `~/.config/rstudio/rstudio-prefs.json` will be automatically generated.

### DIFF
--- a/scripts/default_user.sh
+++ b/scripts/default_user.sh
@@ -12,11 +12,15 @@ else
     addgroup "${DEFAULT_USER}" staff
 
     ## Rocker's default RStudio settings, for better reproducibility
-    mkdir -p "/home/${DEFAULT_USER}/.rstudio/monitored/user-settings"
-    cat <<EOF >"/home/${DEFAULT_USER}/.rstudio/monitored/user-settings/user-settings"
-alwaysSaveHistory="0"
-loadRData="0"
-saveAction="0"
+    mkdir -p "/home/${DEFAULT_USER}/.config/rstudio/"
+    cat <<EOF >"/home/${DEFAULT_USER}/.config/rstudio/rstudio-prefs.json"
+{
+    "save_workspace": "never",
+    "always_save_history": false,
+    "reuse_sessions_for_project_links": true,
+    "posix_terminal_shell": "bash",
+    "initial_working_directory": "/home/${DEFAULT_USER}"
+}
 EOF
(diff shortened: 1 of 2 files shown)


scripts/default_user.sh
```diff
@@ -12,11 +12,15 @@ else
...
-loadRData="0"
-saveAction="0"
+    mkdir -p "/home/${DEFAULT_USER}/.config/rstudio/"
+    cat <<EOF >"/home/${DEFAULT_USER}/.config/rstudio/rstudio-prefs.json"
+{
+    "save_workspace": "never",
+    "always_save_history": false,
+    "reuse_sessions_for_project_links": true,
+    "posix_terminal_shell": "bash",
+    "initial_working_directory": "/home/${DEFAULT_USER}"
```
eitsupi, line 22:
```suggestion
    "posix_terminal_shell": "bash"
```

It is probably safer to delete this line, since writing the user name here would make it impossible to copy this file to another user for use.

PR #535, opened 2022-09-07
Make a sym link of rstudio-server executable

eitsupi:
Create a sym link `/usr/local/bin/rstudio-server` so that there is no need to add `/usr/lib/rstudio-server/bin` to the PATH.


cboettig:
good call!

BEFH:
This has broken the container for me. `rserver` no longer works.

PR #541, opened 2022-09-14
Make a sym link of rserver executable

### LINKED ISSUE #538: rserver executable not found
### Container image name

rocker/rstudio:4.2.1

### Container image digest

_No response_

Linux

### System information

- singularity version 3.7.0-1.el7
- linux 3.10.0-1160.11.1.el7.x86_64

### Bug description

Running the rserver command inside a singulartiy container gives a `FATAL:   "rserver": executable file not found in $PATH` error.

```sh
singularity pull docker://rocker/rstudio:4.2

singularity exec rstudio_4.2.sif rserver
```

Note: The issue does not occurr in `rocker/rstudio:4.1`.

eitsupi:
Close #538

Follow up #535, this change had to be mentioned in the NEWS because it was a significant change for some users.

### DIFF
--- a/scripts/install_rstudio.sh
+++ b/scripts/install_rstudio.sh
@@ -130,8 +130,8 @@ cp /rocker_scripts/pam-helper.sh /usr/lib/rstudio-server/bin/pam-helper
 rm -rf /var/lib/apt/lists/*
 
 # Check the RStudio Server
-echo -e "Check the RStudio Server installation...\n"
-rserver --verify-installation 1
 echo -e "Check the RStudio Server version...\n"
+
 rstudio-server version
+
 echo -e "\nInstall RStudio Server, done!"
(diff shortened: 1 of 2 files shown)


cboettig:
would it be simpler to just add a simlinks for the other binaries there as well?

eitsupi:
> would it be simpler to just add a simlinks for the other binaries there as well?

With the large number of executables as described here, I am wondering if it is necessary to create all the simlinks.
https://github.com/rocker-org/rocker-versioned2/issues/538#issuecomment-1243775174

scripts/install_rstudio.sh
```diff
@@ -128,9 +129,9 @@ cp /rocker_scripts/pam-helper.sh /usr/lib/rstudio-server/bin/pam-helper
 # Clean up
 rm -rf /var/lib/apt/lists/*
 
-# Check the RStudio Server version
+# Check the RStudio Server
+echo -e "Check the RStudio Server installation...\n"
+rserver --verify-installation 1
```
eitsupi, line 134: I am not familiar with the rserver command and am wondering if this is the correct way to do it.
This was used on the following page.
https://cran.r-project.org/bin/linux/fedora/

eitsupi: Since the image size appears to have increased (1743MB -> 1786MB) with the addition of this, it is removed once at beb72a47ffbd6bc5faac45c17288b488ce4fc312.

cboettig (approved):

PR #544, opened 2022-09-19
Update groupmod for already existing group ids within the container

zickzackv:
If the group id already exists the group is not created. 
using the `groupmod` option   "-o, --non-unique", the group and id will be created anyways.

Host: 
   Username: fabian = 1001
   Primary Group: users = 100

Container started with `-e USERID=$(id -u fabian)` and `-e GROUPID=$(id -g fabian)` 
   Username: rstudio = 1001
   Primary Group: users =  100, rstudio = 100


eitsupi (approved):
LGTM.
Thanks for your contribution!

PR #552, opened 2022-10-30
Use `usermod` instead of `addgroup` for adding a group to the user

eitsupi:
Same as rocker-org/rocker#504.

Note that we have already written this in the following line.
https://github.com/rocker-org/rocker-versioned2/blob/fdc8af83524b9b87b37370a1efa3cafe214bc19c/scripts/init_userconf.sh#L66


eitsupi:
Merge as it is a minor change and passes CI.

Issue #543, opened 2022-09-15
Report image size of test builds in CI logs


eitsupi:
Currently we have to download the artifact to see the built image size, but this is a hassle.
The image size is listed in the json file and can be easily extracted with jq.

mskyttner:
Sorry if this is not 100% related to this issue, but regarding image sizes I noticed that the size has almost doubled for the rocker/ml-verse image when comparing 4.2.1 to 4.2.2 (which is on jammy, I believe):

```bash
$ docker images | grep rocker/ml-verse | head -n 2

rocker/ml-verse                          4.2.2                          7723b958405a   5 days ago      13.2GB
rocker/ml-verse                          4.2.1                          dcfdabe76d04   6 weeks ago     6.6GB

```
Is this increased size in the image expected/normal? I was just surprised that switching to jammy would mean so much for the image size?

eitsupi:
@mskyttner It is entirely expected due to the base image switch and has nothing to do with this issue.

Compare the following two
https://github.com/rocker-org/rocker-versioned2/wiki/cuda_d107b047f914
https://github.com/rocker-org/rocker-versioned2/wiki/cuda_d9acea13af6f

Perhaps more information should have been included in NEWS.md about switching base images.

PR #516, opened 2022-08-17
Bump cuda version and drop cuda version tags for `ubuntu:jammy` based images

eitsupi:
close #282, close #503

We have maintained two `rocker/cuda` variants, cuda10.1 and cuda11.1, but now that the base image will be `ubuntu:jammy`, we will unify the cuda variants into one, as we can no longer use anything other than cuda11.7.X.

1. cuda version tags, like `rocker/cuda:cuda10.1`, `rocker/cuda:4.2.1-cuda10.1` will be obsolete for R 4.2.2 or later images.
2. The aliases of `rocker/cuda` which was pushed to `rocker/r-ver` (`rocker/r-ver:X.Y.Z-cuda10.1` and `rocker/r-ver:X.Y.Z-cuda11.1`) will be obsolete.

Only the R version tags like `rocker/cuda:4.2.2`, `rocker/cuda:latest` will remain.

Note: This Pull Request can only be merged after the R 4.2.2 release.

**ToDo**

- [x] Update `stacks/4.2.1.json` by hand.
  - [x] Remove `latest` tags.
  - [x] Change CRAN and CTAN urls.
- [x] Run `./build/make-stacks.R` -> `make clean` -> `make setup`
- [x] Update NEWS.md
- [ ] Update the reference page

PR #562, opened 2022-11-09
Update NEWS about cuda, ml images R >= 4.2.2 (`ubuntu:jammy` based)

eitsupi:
Since there was not enough description in the news (about #516), I added it.
c.f. https://github.com/rocker-org/rocker-versioned2/issues/543#issuecomment-1307707608

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,8 +14,8 @@
 - cuda version tags, like `rocker/cuda:cuda10.1`, `rocker/ml:4.2.1-cuda10.1` are obsolete for R 4.2.2 or later images.
   Only the R version tags like `rocker/cuda:4.2.2`, `rocker/ml-verse:latest` will remain.
   `rocker/cuda` R >= 4.2.2 corresponds to `rocker/cuda:cuda11.1` before R.4.2.1 and
-  uses `nvidia/cuda` as the base image.
-  Therefore, the image size is significantly different from `rocker/cuda` R <= 4.2.1,
+  uses `nvidia/cuda:X.Y.Z-cudnn8-devel-ubuntu22.04` as the base image.
+  Therefore, the image size is significantly larger (+6GB) from `rocker/cuda` R <= 4.2.1,
   which installed cuda on `rocker/r-ver` (based on `ubuntu:focal`).
   ([#516](https://github.com/rocker-org/rocker-versioned2/pull/516))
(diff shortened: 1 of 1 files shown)


eitsupi:
@cboettig Could you take a look at this?

NEWS.md
```diff
@@ -11,10 +11,16 @@
 
 ### Changes in pre-built images
 
-- cuda version tags, like `rocker/cuda:cuda10.1`, `rocker/cuda:4.2.1-cuda10.1` are obsolete for R 4.2.2 or later images.
-  Only the R version tags like `rocker/cuda:4.2.2`, `rocker/cuda:latest` will remain.
+- cuda version tags, like `rocker/cuda:cuda10.1`, `rocker/ml:4.2.1-cuda10.1` are obsolete for R 4.2.2 or later images.
+  Only the R version tags like `rocker/cuda:4.2.2`, `rocker/ml-verse:latest` will remain.
+  `rocker/cuda` R >= 4.2.2 corresponds to `rocker/cuda:cuda11.1` before R.4.2.1 and
+  uses `nvidia/cuda` as the base image.
+  Therefore, the image size is significantly different from `rocker/cuda` R <= 4.2.1,
```
cboettig, line 18: I think we should say "significantly larger" rather than "different".

cboettig: also, maybe we should comment specifically that it uses `nvidia/cuda` devel images (tag `cudnn8-devel-ubuntu22.04`) -- arguably we might get away with `runtime`, I'm not sure we need devel and could save some significant space.  (On the plus side users compiling against cuda libs might appreciate this?)

cboettig (approved):
Looks great, thanks! minor comments here.

NEWS.md
eitsupi: Thanks for the review. I have made the changes.

BTW, `nvidia/cuda:11.8.0-cudnn8-runtime-ubuntu22.04` seems to be half the size of `nvidia/cuda:11.8.0-cudnn8-devel-ubuntu22.04`.
https://hub.docker.com/r/nvidia/cuda/tags?page=1&name=22.04

I don't usually use cuda so I don't know what is different about these images and how they benefit the user.
It may be worth creating an issue about changing the base image.

cboettig (approved):

PR #564, opened 2022-11-13
GDAL & GEOS should use cmake, closes #458

### LINKED ISSUE #458: GEOS is already cmake- only and GDAL 3.5.0 optionally, from 3.6.0 cmake only
Please note that GEOS is already cmake- only, as PROJ. GDAL 3.5.0 optionally, from 3.6.0 cmake only. If revising anyway, maybe make the cmake changes at the same time ?

_Originally posted by @rsbivand in https://github.com/rocker-org/rocker-versioned2/pull/455#discussion_r879578199_

cboettig:


scripts/experimental/install_dev_osgeo.sh
```diff
@@ -71,16 +71,16 @@ LD_LIBRARY_PATH=/usr/local/lib:$LD_LIBRARY_PATH
 if [ "$GEOS_VERSION" = "latest" ]; then
     GEOS_VERSION=$(wget -qO- "https://api.github.com/repos/libgeos/geos/git/refs/tags" | grep -oP "(?<=\"ref\":\s\"refs/tags/)\d+\.\d+\.\d+" | tail -n -1)
 fi
-
-wget http://download.osgeo.org/geos/geos-"${GEOS_VERSION}".tar.bz2
+wget -q http://download.osgeo.org/geos/geos-${GEOS_VERSION}.tar.bz2
```
eitsupi, line 74: Is this change necessary? (I believe it needs to be enclosed in at least double quotes)

scripts/experimental/install_dev_osgeo.sh
```diff
@@ -71,16 +71,16 @@ LD_LIBRARY_PATH=/usr/local/lib:$LD_LIBRARY_PATH
...
 bzip2 -d geos-*bz2
 tar xf geos*tar
-rm geos*tar
```
eitsupi, line 78: If I remove this line, won't the tar file remain?

scripts/experimental/install_dev_osgeo.sh
```diff
@@ -113,12 +113,12 @@ fi
 
 wget "$GDAL_DL_URL" -O gdal.tar.gz
 tar -xf gdal.tar.gz
-rm gdal.tar.gz
```
eitsupi, line 116: Is this line no longer needed?

eitsupi (changes requested):
Sorry, but I have not been able to do any local build testing, I think I should add it to CI, but am afraid to do so because it takes too long to build.

cboettig:
@eitsupi looks like linter is now happy and it passes the simple test of being able run the builder without error on rocker/r-ver (see https://github.com/rocker-org/rocker-versioned2/actions/runs/3458693351) which takes about an hour to compile all three elements from source.  

I had to tweak the helper function for detecting `latest` download blobs to avoid getting two URL hits at once, which caused an error.  

Let me know what you think & thanks again for the review & suggestions!

eitsupi (approved):
Thank you for working on this.
Sorry I can't give you a sufficient review as I understand very little about the build, but thanks to your addition of CI it seems to be working fine!

I modified the CI definition to be the same as the other CI definitions and modified the scripts slightly.

scripts/experimental/install_dev_osgeo.sh
```diff
@@ -26,7 +26,7 @@ function apt_install() {
 }
 
 function url_latest_gh_released_asset() {
-    wget -qO- "https://api.github.com/repos/$1/releases/latest" | grep -oP "(?<=\"browser_download_url\":\s\")https.*\.tar.gz(?=\")"
+    wget -qO- "https://api.github.com/repos/$1/releases/latest" | grep -oP "(?<=\"browser_download_url\":\s\")https.*\.tar.gz(?=\")" | grep -v "test"
```
eitsupi, line 29: It seems to me that using the string `test` here is not a good way to convey the intent, so `head -n 1` seems a better choice.

PR #568, opened 2022-11-30
Set WORKDIR for *all* images

yuvipanda:
Without this, trying to run these for anything other than RStudio puts us in `/` rather than in `/home/rstudio`.

```bash
$ docker run -it rocker/geospatial:4.2.2 pwd
/
`

RStudio puts the user in /home/rstudio as expected (probably something
to do with PAM, I assume). This change makes the behavior of using the
images without RStudio match that of using it with.

### DIFF
--- a/stacks/devel.json
+++ b/stacks/devel.json
@@ -37,7 +37,8 @@
       },
       "COPY": "scripts /rocker_scripts",
       "RUN": "/rocker_scripts/setup_R.sh",
-      "CMD": "[\"R\"]"
+      "CMD": "[\"R\"]",
+      "WORKDIR": "/home/rstudio"
     },
     {
       "IMAGE": "rstudio",
@@ -59,6 +60,7 @@
         "/rocker_scripts/install_quarto.sh"
       ],
       "CMD": "[\"/init\"]",
+      "WORKDIR": "/home/rstudio",
       "EXPOSE": 8787
     },
     {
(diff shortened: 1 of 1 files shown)


yuvipanda:
I'm also working on an (unpublished! do not spread too widely :D) opinionated guide to making JupyterHub images, and recommend using rocker as the base for images that will be primarily used with R. I'm having folks inherit from geospatial, and this reduces an extra section I've to set there :) https://jupyterhub-image.guide/rocker.html#tell-jupyter-where-to-start

eitsupi:
Is there any reason why we cannot simply use the `--workdir` option of the `docker run` command?

yuvipanda:
@eitsupi the equivalent of that in kubernetes is what I have documented it to be. However, it does mean that everyone has to remember to do that, or the default working directory is something the default user can not write into. This is different from the experience you get when you run the rstudio part of the image. So my hope is that for end users of the image, there are fewer surprises on why something that works when you use rstudio doesn't when you don't.

Is there a strong reason to *not* set workdir in the image?

cboettig:
Note that the default user on the image is also root and thus can still write to the default workdir -- and these settings for default user and default workdir are just identical to the base Ubuntu image (and most other base OS images, right?).  We've generally these standard defaults, overriding them may have some small chance of breaking some downstream user.

The default application is to launch the S6 init system, which brings up rstudio server, which recognizes the standard multi-user environment and allows each user to log in with their linux user name and password (i.e. as set by `adduser`).  

I gather this design is a bit at odds with the more modern kubernetes convention for Docker, where a container is assumed to have only a single user configured.  (Just like our use of an init system is a bit at odds with the convention that a container should run a single application).

yuvipanda:
If you're worried this might be a breaking change for some users, happy to close the PR and let things be as they are.

Thank you for all your work on the project!

eitsupi:
Ok, I will close this for now. Thanks!

PR #575, opened 2022-12-09
Add lsb-release for shiny-server

eitsupi:
As a result of `lsb-release` being removed from `rocker/r-ver` in #511, it appears that `install_shiny_server.sh` is no longer functional.
We need to install `lsb-release` for shiny-server.


cboettig:
ha, good catch.  LGTM, merge when ready

eitsupi:
I missed this in #511 because a test case was missing, so I added a test case in c2379fe.

PR #576, opened 2022-12-10
Remove installation of wgrib from geospatial builds

rdenham:
Gdal handles grib2 files well so no longer seems to be a need to install the 
[wgrib](https://www.cpc.ncep.noaa.gov/products/wesley/wgrib2/) utility, which
is used to read and write grib2 files.


cboettig:
Thanks @rdenham , this is a good change and will reduce bloat.  

@eitsupi I think you've already resolved the issue wit the one failing test (lsb_release for shiny) elsewhere and so this should be good to merge!

eitsupi:
Yes, the test failure is nothing to do with this PR.

I am only wondering if removing this will not change the behavior on focal based images as well.

cboettig:
@eitsupi good call.  For consistency I agree we should leave wgrib on the `focal` images to meet reproducibility expectations (though maybe we should have had a higher bar of criteria to add custom installs like this that don't go through apt or RSPM for that reason).  

Do you have a preferred pattern for checking the release version in a shell script?  I think looking at `/etc/issue` is reasonable (has the release number but not the codename, but I think avoids any additional dependency like `lsb_release -c` ...

@rdenham could you add a conditional instead of removing the wgrib install such that it only runs on 20.04?  (Note that the latest release of this stack and going forward are now all on 22.04, so it's a good time to make a change like this).

eitsupi:
> Do you have a preferred pattern for checking the release version in a shell script? I think looking at `/etc/issue` is reasonable (has the release number but not the codename, but I think avoids any additional dependency like `lsb_release -c` ...

I recommend like this:

```sh
# shellcheck source=/dev/null
source /etc/os-release
if [ "${UBUNTU_CODENAME}" == "focal" ]; then
    /rocker_scripts/install_wgrib2.sh
fi
```

rdenham:
> @rdenham could you add a conditional instead of removing the wgrib install such that it only runs on 20.04? (Note that the latest release of this stack and going forward are now all on 22.04, so it's a good time to make a change like this).

Sure. Do you think there is merit in also removing unnecessary files left over from the install as well, or just leave it as is given it will only affect older builds?

rdenham:
> > @rdenham could you add a conditional instead of removing the wgrib install such that it only runs on 20.04? (Note that the latest release of this stack and going forward are now all on 22.04, so it's a good time to make a change like this).
> 
> Sure. Do you think there is merit in also removing unnecessary files left over from the install as well, or just leave it as is given it will only affect older builds?

Just replying to myself, to get this over the line, I think we should just use the wgrib install script as is for the focal images, without further modification.

eitsupi:
> Do you think there is merit in also removing unnecessary files left over from the install as well, or just leave it as is given it will only affect older builds?

I think it would be great if you could add a process to delete unnecessary files.
The focal-based images will continue to be built.

cboettig (approved):

PR #578, opened 2022-12-16
Allow building `rocker/rstudio:latest-daily` on linux/arm64 platform with skipping install quarto CLI

eitsupi:
Related to #144 

Skip installation of quarto-cli to allow arm64 build of `rocker/rstudio:latest-daily`.
(workflow setup will be a following PR)


cboettig:
:tada:

Issue #557, opened 2022-11-02
Failed to locate timezone database in ubuntu:jammy based images (`rocker/r-ver:4.2.2` etc.)


VincentGuyader:
### Container image name

rocker/verse

### Container image digest

rocker/verse@sha256:b8bb73f75e130e5767d64850327971d8cade933171e0a68460e1444b1c106471

### What operating system are you seeing the problem on?

Linux

### System information

 Docker Desktop 4.11.0 (83626)

### Bug description

timezone database is missing in rocker/verse

### How to reproduce this bug?

```sh
docker run -it rocker/verse@sha256:b8bb73f75e130e5767d64850327971d8cade933171e0a68460e1444b1c106471 bash
R

library(lubridate)
> Warning message:
>  Failed to locate timezone database

OlsonNames()
> character(0)
```

eitsupi:
> The build action for 4.2.2 [shows that tzdata is removed](https://github.com/rocker-org/rocker-versioned2/actions/runs/3360364550/jobs/5569477110#step:7:9807) ([here as well](https://github.com/rocker-org/rocker-versioned2/actions/runs/3360364550/jobs/5569477110#step:7:14274)).

@dpprdan Thanks for investigating this!

dpprdan:
@VincentGuyader The root cause seems to be that the tzdata apt package is not installed (see #558)

Surprisingly, tzdata _is_ listed as installed in the [wiki for the latest 4.2.2 image](https://github.com/rocker-org/rocker-versioned2/wiki/r-ver_e2a4fc8df574#apt-packages).

eitsupi:
🤔

```sh
$ dpkg -s tzdata
Package: tzdata
Status: deinstall ok config-files
Priority: required
Section: localization
Installed-Size: 3772
Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
Architecture: all
Multi-Arch: foreign
Version: 2022e-0ubuntu0.22.04.0
Config-Version: 2022e-0ubuntu0.22.04.0
Provides: tzdata-bookworm
Depends: debconf (>= 0.5) | debconf-2.0
Description: time zone and daylight-saving time data
 This package contains data required for the implementation of
 standard local time for many representative locations around the
 globe. It is updated periodically to reflect changes made by
 political bodies to time zone boundaries, UTC offsets, and
 daylight-saving rules.
Homepage: https://www.iana.org/time-zones
[...]

PR #583, opened 2022-12-21
[CI] Deinstalled apt packages should not be listed in reports

eitsupi:
Reported in https://github.com/rocker-org/rocker-versioned2/issues/557#issuecomment-1300173591

I also modify the workflow so that when the report template is updated, it will be tested on the PR.


eitsupi:
It seems working fine on the test run, so I'll merge.

PR #624, opened 2023-03-29
[CI] Reduce building images (supporting R versions)

### LINKED ISSUE #614: Support Term?
With the release of R 4.2.3, we finally have more versions built than the number of GitHub Actions parallel execution workflows.

I don't think it is a sustainable approach to keep increasing the number of builds, so I am wondering if it is possible to stop building images that are considered unimportant (e.g. not the final version of X.Y like R 4.0.1).

image

eitsupi:
Resolve #614

Reduce the number of images to be built according to the following new standard:

- All patch versions associated with the last 2 minor versions (release and oldrel)
- All minor versions R >= 4.0


eitsupi:
@cboettig Is it ok to merge this?
Stop building all versions from the next build (when R 4.3.0 released).

eddelbuettel:
Why not keep it simpler and just use the final version in each major.minor pair?

```r
"r_version": ["4.0.5", "4.1.3",  "4.2.3"],
```

eitsupi:
> Why not keep it simpler and just use the final version in each major.minor pair?

I assume that there is a demand to stay with a specific patch version for issues with versions other than R itself (R package, RStudio, etc.).

However, I do not have a strong opinion.
The last two patch versions (required for CRAN date freezing) and each minor version are fine by me.

```json
"r_version": ["4.0.5", "4.1.3", "4.2.2", "4.2.3"],
```

cboettig:
LGTM!

PR #628, opened 2023-04-01
Use `jupyter lab` instead of `jupyter notebook` as default command of `rocker/binder`

### LINKED ISSUE #621: rocker/binder not opening Rstudio in Jupyter?
Hi all --- I may be missing something obvious (or simply be slightly out of my depth), so apologies if this has a trivial explanation/solution...

But: up until a while ago (I *think* that's preceding the recent upgrade of [Jupyter to Notebook 7](https://jupyter-notebook.readthedocs.io/en/latest/migrate_to_notebook7.html), I was able to build a local docker container based on `rocker/binder` and then run it through a browser opening an `Rstudio` instance, something like
```
docker run --rm -it [OTHER OPTIONS] my-container
```
For a while now, this operation fails. I can still use `mybinder` to run the container via GitHub, but the launching it locally and aiming to go on the `Rstudio` browser fails. I can point my browser to the application (typically that's `http://127.0.0.1:8888/tree`) but there's no option for me to launch `Rstudio` --- in fact, there's an option to run `R`, but it fails, giving a message like this:
image

I also get a message 
```
UPDATE Read [the migration plan](https://jupyter-notebook.readthedocs.io/en/latest/migrate_to_notebook7.html) to Notebook 7 to learn about the new features and the actions to take if you are using extensions - Please note that updating to Notebook 7 might break some of your extensions.
```
which links to a page that seems to suggest that Jupyter 7 will break things spectacularly... 

Like I said, it may be that I don't know enough about Jupyter and there's in fact a very easy way around it, but I wonder whether others have encountered this and/or it is in fact an impossibility to run `Rstudio` from `rocker/binder`, locally, now?

Thanks for your help!

eitsupi:
Fix #621

Jupyter notebook no longer seems to work well.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,12 @@
 # News
 
+## 2023-04
+
+### Changes in pre-built images
+
+- `rocker/binder`'s default command is changed from `jupyter notebook` to `jupyter lab`.
+  ([#628](https://github.com/rocker-org/rocker-versioned2/pull/628))
+
 ## 2022-12
 
 ### Changes in pre-built images
--- a/dockerfiles/binder_4.0.0.Dockerfile
+++ b/dockerfiles/binder_4.0.0.Dockerfile
@@ -11,7 +11,7 @@ RUN /rocker_scripts/install_jupyter.sh
 
 EXPOSE 8888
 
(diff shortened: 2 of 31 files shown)


cboettig (approved):
excellent, yes I think this is the way to go.

PR #633, opened 2023-04-19
add geoparquet/arrow drivers for GDAL

cboettig:


cboettig:
@eitsupi can you take a look over this when you get a chance?  

Apologies about the linter error, not quite sure what is up there.  I'm basically following the syntax from the [official arrow installation instructions](https://arrow.apache.org/install/), though I'm not quite clear why they have gone about this by using wget + a `.deb` file install in order to update `sources.list.d` dir rather than the traditional PPA route...

cboettig:
@mdsumner would you mind taking a look over this updated gdal build recipe?

scripts/experimental/install_dev_osgeo.sh
```diff
@@ -6,7 +6,7 @@ set -e
 ## 'latest' means installing the latest release version.
 
 ## build ARGs
-NCPUS=${NCPUS:-"-1"}
+NCPUS=${NCPUS:-"1"}
```
eitsupi, line 9: I thought this was for `install2.r`.
Is there any reason to use this with cmake?
Is there a problem with the default value of cmake?

scripts/experimental/install_dev_osgeo.sh
```diff
@@ -64,6 +64,11 @@ apt_install \
     cmake \
     libtiff5-dev
 
+## geoparquet support
+wget https://apache.jfrog.io/artifactory/arrow/"$(lsb_release --id --short | tr 'A-Z' 'a-z')"/apache-arrow-apt-source-latest-"$(lsb_release --codename --short)".deb
+apt_install -y -V ./apache-arrow-apt-source-latest-"$(lsb_release --codename --short)".deb
+apt-get update && apt-get install -y -V libarrow-dev  libparquet-dev libarrow-dataset-dev
```
eitsupi, line 70:
```suggestion
apt-get update && apt-get install -y -V libarrow-dev libparquet-dev libarrow-dataset-dev
```

scripts/experimental/install_dev_osgeo.sh
```diff
@@ -64,6 +64,11 @@ apt_install \
     cmake \
     libtiff5-dev
 
+## geoparquet support
+wget https://apache.jfrog.io/artifactory/arrow/"$(lsb_release --id --short | tr 'A-Z' 'a-z')"/apache-arrow-apt-source-latest-"$(lsb_release --codename --short)".deb
```
eitsupi, line 68:
```suggestion
wget https://apache.jfrog.io/artifactory/arrow/"$(lsb_release --id --short | tr [:upper:] [:lower:])"/apache-arrow-apt-source-latest-"$(lsb_release --codename --short)".deb
```

eitsupi:
(I closed and reopened it once to trigger CI)

scripts/experimental/install_dev_osgeo.sh
```diff
@@ -64,6 +64,11 @@ apt_install \
     cmake \
     libtiff5-dev
 
+## geoparquet support
+wget https://apache.jfrog.io/artifactory/arrow/"$(lsb_release --id --short | tr [:upper:] [:lower:])"/apache-arrow-apt-source-latest-"$(lsb_release --codename --short)".deb
```
eitsupi, line 68:
```suggestion
wget https://apache.jfrog.io/artifactory/arrow/"$(lsb_release --id --short | tr '[:upper:]' '[:lower:]')"/apache-arrow-apt-source-latest-"$(lsb_release --codename --short)".deb
```

scripts/experimental/install_dev_osgeo.sh
```diff
@@ -33,6 +33,7 @@ export DEBIAN_FRONTEND=noninteractive
 apt_install \
     gdb \
     git \
+    lsb_release \
```
eitsupi, line 36:
```suggestion
    lsb-release \
```

scripts/experimental/install_dev_osgeo.sh
cboettig: Right, I think this was initially introduced for `install2.r`, but compiling OSGEO C libraries from source is perhaps even more time consuming, and definitely benefits from parallelization.  `cmake` errors with the default value of `-1`.  

I suppose we could use separate env vars to define the number of cores available to `install2.r` and the number of cores available to `cmake`, but I suspect that would merely confuse users and developers more?  

Maybe it would be better to auto-detect available cores? Could be more fragile though.

cboettig:
:heart:  Thanks so much for the help here @eitsupi, looks great!  If you're ok with the `NCPUS` change here feel free to merge when ready.

scripts/experimental/install_dev_osgeo.sh
eitsupi: I am not sure if this is used by the users. Is there any use case other than `-1` to install2.r?

I disagree with the changing default value to `1` as it takes longer to run on CI.
How about having the command get the number of available cores if `-1` is set?

cboettig: Yes, that's a very good point.  

I guess

```
nproc --all
```

is the best way to do that?

eitsupi: > I guess
> 
> ```
> nproc --all
> ```
> 
> is the best way to do that?

I am not familiar with the commands to do this but it seems good!

cboettig: `nproc` part of coreutils in linux, so should be pretty widely available.  I've pushed a change for this now (below); though no doubt the linter won't like my bash :-)

PR #642, opened 2023-04-23
ci(reports): fix report rename mechanism

### LINKED ISSUE #637: Fix wiki home
After #624, the wiki home build seems fail.

eitsupi:
Fix #637


eitsupi:
It seems working, so I try merging this.

The CI error is due to super-linter's bug of R linting. (Maybe we should switch to MegaLinter (#643))

Issue #582, opened 2022-12-19
Supportin multiple CUDA versions?  (CUDA bumps to 11.8 on the 4.2.2 images)


cboettig:
Just wondering if we want to revisit support for multiple CUDA tags across a given/latest version of R.  We bumped up to 11.8 with the R 4.2.2 / ubuntu 22.04 release, and I'm observing that it is not compatible with host platforms that might be running older CUDA drivers.  (note that the host machine has to have driver versions greater than or equal to the libraries on the containers).  

NVIDIA provides 11.7.0 and 11.7.1 on ubuntu-22, as well as 11.8.0 which we're using.  

weirdly, I have one machine with NVIDIA Driver  470.141.03   CUDA Version: 11.4, which runs the 11.8 image fine, but a machine with slightly newer drivers :  Driver Version: 515.65.01    CUDA Version: 11.7, can only run 11.7.1 but not the 11.8 dockerfiles.  

welcome other experiences, I'll try and triangulate this one a bit more too.

hute37:
I'm busy in CUDA (keras/tensorflow) setup for aged *NVIDIA Tesla K80 Azure Datacenter GPU* (2014)

* [NC-series](https://learn.microsoft.com/en-us/azure/virtual-machines/nc-series)

These GPU are rather old and near to be [dismissed by Microsoft](https://learn.microsoft.com/en-us/azure/virtual-machines/nc-series-retirement), but very cheap, useful for educational vm, used by students in our institution.

The NVIDIA kernel driver supported is branch *470*:

in `rocker/ml:4.2.3` container:

```
| NVIDIA-SMI 470.182.03   Driver Version: 470.182.03   CUDA Version: 11.8     |
```

from docs, 470 should be supported in all CUDA-11.x releases

* [[3.3. Deployment Considerations for Forward Compatibility](https://docs.nvidia.com/deploy/cuda-compatibility/index.html#deployment-consideration-forward)](https://docs.nvidia.com/deploy/cuda-compatibility/index.html#binary-compatibility__table-toolkit-driver)


But comparing repositories, the 470 (user-mode) driver is missing

* [ubuntu 22.04](https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2204/x86_64/):

```
cuda-compat-11-7_515.43.04-1_amd64.deb
cuda-compat-11-7_515.48.07-1_amd64.deb
cuda-compat-11-7_515.65.01-1_amd64.deb
[...]

PR #657, opened 2023-06-01
Only switch to libblas on focal

### LINKED ISSUE #471: matplotlib plots crash due to conflicting BLAS versions in numpy
The pre-compiled binaries for numpy are built against a different BLAS library than the openblas library included on rocker images.  This leads to unexpected errors unless numpy is installed from source, e.g. 

```
pip3 install --no-binary="numpy" numpy --ignore-installed
```

I don't yet know the best work around here, it's difficult as the error is both relatively opaque for users to track down, and there are just oh so many ways users will install and re-install numpy from binary without realizing it in different virtualenvs.  

A start but no means a complete solution would be to at least build numpy from source in the python images.  Open to any better solutions / ideas?  (cc @yuvipanda ?)

eitsupi:
Close #471 (and related to https://github.com/rocker-org/rocker-versioned2/issues/582#issuecomment-1570584990)

This workaround seems to be sufficient if it is executed only on Ubuntu 20.04, since OpenBLAS on Ubuntu 22.04 does not seem to have the problem of crashing numpy.

### DIFF
--- a/scripts/install_python.sh
+++ b/scripts/install_python.sh
@@ -40,13 +40,16 @@ rm -rf /tmp/downloaded_packages
 ## https://github.com/rocker-org/rocker-versioned2/issues/340
 strip /usr/local/lib/R/site-library/*/libs/*.so
 
-## Don't use OpenBLAS with reticulate
+## Don't use OpenBLAS with reticulate on Ubuntu 20.04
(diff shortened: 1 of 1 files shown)


cboettig (approved):
awesome, thanks for this!

PR #668, opened 2023-06-20
Install JupyterHub as part of install_jupyter.sh

yuvipanda:
This allows using the built image with JupyterHubs for running RStudio.

### DIFF
--- a/scripts/install_jupyter.sh
+++ b/scripts/install_jupyter.sh
@@ -29,7 +29,7 @@ fi
 # install python
 /rocker_scripts/install_python.sh
 
-python3 -m pip install --no-cache-dir jupyter-rsession-proxy notebook jupyterlab
+python3 -m pip install --no-cache-dir jupyter-rsession-proxy notebook jupyterlab jupyterhub
 
 install2.r --error --skipmissing --skipinstalled -n "$NCPUS" remotes
 


eitsupi:
Thanks for working on this!

The CI failure seems nothing to do with this change.

eitsupi:
(The failed test was fixed in #669)

yuvipanda:
Great! I've rebased on latest...

cboettig:
nice!

yuvipanda:
@cboettig do you know when this might make its way to Dockerhub? I see in https://hub.docker.com/r/rocker/binder/tags that the last push was a few days before this PR got merged.

cboettig:
@yuvipanda thanks for the ping! looks like the most recent [automated build](https://github.com/rocker-org/rocker-versioned2/actions/workflows/core.yml) failed, I've tickled a re-build.

yuvipanda:
yay, thanks @cboettig! That works well :)

fwiw, we're starting to recommend people just use the upstream rocker/binder image on 2i2c JupyterHubs when they want to use rstudio (https://github.com/2i2c-org/infrastructure/pull/2726) and thought you might like to know :)

eitsupi:
> @yuvipanda thanks for the ping! looks like the most recent [automated build](https://github.com/rocker-org/rocker-versioned2/actions/workflows/core.yml) failed, I've tickled a re-build.

Indeed, the reason it was not deployed was not because it did not fail the auto-build, but because this change did not trigger the auto-build in the first place.

Currently, only R or RStudio version upgrades trigger the auto-build, and other builds are triggered manually about once a month on a whim when I see an update to the base image (`ubuntu`).
I really would like to automatically detect updates to the base image, but have not been able to do so. (Although Dependabot can do it, so perhaps there is a way.)

cboettig:
Thanks @eitsupi .  I was wondering if we should add a cron trigger to the action to ensure an auto-build of the latest image on a predictable schedule?  Maybe once a week?  I think it would be nice to ensure that the `latest` image is a relatively recent build across the board, e.g. that we're not a month behind pulling in patches to security-related libraries that aren't installed in the base image by default, (e.g. network-related libs like openssl / curl). This also serves to keep R packages more up-to-date though that's probably less important.

yuvipanda:
@cboettig yep, cron is what we do in jupyter/docker-stacks (https://github.com/jupyter/docker-stacks/blob/df5d516e34efafc699783008c007cfcc691b164c/.github/workflows/docker.yml#L4)

eddelbuettel:
I do it too in some GH Actions.  They tend to send nastygrams if the repos were commit-less for sixty days so for rocker's r-devel, drd, drp, san I also commit the output of `date > timestamp.txt` in the action which, while silly, keeps the nastygrams at bay

yuvipanda:
@cboettig @eitsupi I opened https://github.com/rocker-org/rocker-versioned2/pull/671!

eitsupi:
I have created a new issue #672.

Issue #670, opened 2023-06-21
Using pip without venv or conda env will stop working soon


yuvipanda:
https://peps.python.org/pep-0668/ was adopted 2 years ago, and it's finally rolling around to distros downstream. The upshot is that just using `pip` with system python will **no longer work**. With the recently released ubuntu:23.04 (or debian:bookworm), if you try this:

```Dockerfile
FROM ubuntu:23.04
RUN apt-get update && apt-get -y install python3 python3-pip
RUN pip install flask
```

You'll get the following error:

```
 > [3/3] RUN pip install flask:                                                     #6 4.003 error: externally-managed-environment                                      #6 4.003                                                                            #6 4.003 × This environment is externally managed                                   #6 4.003 ╰─> To install Python packages system-wide, try apt install                #6 4.003     python3-xyz, where xyz is the package you are trying to
#6 4.003     install.
#6 4.003     
#6 4.003     If you wish to install a non-Debian-packaged Python package,
#6 4.003     create a virtual environment using python3 -m venv path/to/venv.
#6 4.003     Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
#6 4.003     sure you have python3-full installed.
#6 4.003     
#6 4.003     If you wish to install a non-Debian packaged Python application,
#6 4.003     it may be easiest to use pipx install xyz, which will manage a
#6 4.003     virtual environment for you. Make sure you have pipx installed.
#6 4.003     
#6 4.003     See /usr/share/doc/python3.11/README.venv for more information.
#6 4.003 
#6 4.003 note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
#6 4.003 hint: See PEP 668 for the detailed specification.
------
```

https://pythonspeed.com/articles/externally-managed-environment-pep-668/ has more useful information. 

The `install_jupyter.sh` script currently does this (https://github.com/rocker-org/rocker-versioned2/blob/a00b24f7916648b1a383f0da49974fe05b0783ec/scripts/install_jupyter.sh#L32), and hence will stop working once we bump to the newer version of Ubuntu. 

There are two options going forward to prevent breaking:

1. Switch to using a venv, and install stuff inside that.
[...]

PR #703, opened 2023-09-22
numpy in  install_dev_osgeo.sh

mdsumner:
I'm not entirely sure about this, and maybe setuptools is not also required - but osgeo.gdal and family require numpy, so this gets pip and installs it before building gdal

### DIFF
--- a/scripts/experimental/install_dev_osgeo.sh
+++ b/scripts/experimental/install_dev_osgeo.sh
@@ -134,6 +134,7 @@ fi
 ## ensure that osgeo.gdal and family get installed for python use https://github.com/OSGeo/gdal/blob/master/swig/python/README.rst
 wget https://bootstrap.pypa.io/get-pip.py
 python3 get-pip.py
+rm get-pip.py
 pip install numpy setuptools
 
 wget "$GDAL_DL_URL" -O gdal.tar.gz
--- a/scripts/experimental/install_dev_osgeo.sh
+++ b/scripts/experimental/install_dev_osgeo.sh
@@ -131,6 +131,11 @@ else
     GDAL_DL_URL="https://download.osgeo.org/gdal/${GDAL_VERSION}/gdal-${GDAL_VERSION}.tar.gz"
 fi
 
+## ensure that osgeo.gdal and family get installed for python use https://github.com/OSGeo/gdal/blob/master/swig/python/README.rst
+wget https://bootstrap.pypa.io/get-pip.py
+python3 get-pip.py
+pip install numpy setuptools
+
 wget "$GDAL_DL_URL" -O gdal.tar.gz
 tar -xf gdal.tar.gz
 rm gdal*tar.gz
--- a/scripts/experimental/install_dev_osgeo.sh
+++ b/scripts/experimental/install_dev_osgeo.sh
@@ -135,7 +135,7 @@ fi
 wget https://bootstrap.pypa.io/get-pip.py
 python3 get-pip.py
 rm get-pip.py
-pip install numpy setuptools
+pip install  --no-cache-dir numpy setuptools
 
 wget "$GDAL_DL_URL" -O gdal.tar.gz
 tar -xf gdal.tar.gz


scripts/experimental/install_dev_osgeo.sh
```diff
@@ -131,6 +131,11 @@ else
     GDAL_DL_URL="https://download.osgeo.org/gdal/${GDAL_VERSION}/gdal-${GDAL_VERSION}.tar.gz"
 fi
 
+## ensure that osgeo.gdal and family get installed for python use https://github.com/OSGeo/gdal/blob/master/swig/python/README.rst
+wget https://bootstrap.pypa.io/get-pip.py
+python3 get-pip.py
+pip install numpy setuptools
```
eitsupi, line 137: I don't see the need for this, but could you at least change the install command to not leave the cache in the layer and delete what wget downloaded?

https://github.com/rocker-org/rocker-versioned2/blob/5fa99b369510466ea8a2803d29e3417f5212a0fc/scripts/install_jupyter.sh#L32

eitsupi (changes requested):

scripts/experimental/install_dev_osgeo.sh
mdsumner: yes sorry, bit out of my depth I think, I'll keep exploring this and the other python pathway

the need is to have osgeo.gdal available on the same dev-osgeo basis, and callable from R

eitsupi: I understand what the process you wrote is trying to do. For consistency's sake, I'd like pip not to be installed this way.
How about using the `install_python.sh` script?

mdsumner: oh thanks yes! will do , never before understood what *should* be done because the official-seeming docs only show ensure-pip and this weird script way

cboettig:
Hi @mdsumner @eitsupi 

Thanks both for this!  Yes, agree with @eitsupi it would be nice to build on the `install_python.sh` setup (or at least match the pip install strategy there which installs `python3-pip` from the Ubuntu LTS repos. Generally the philosophy here is to grab external dependencies where possible from Ubuntu LTS repos, which are maintained for the lifetime of the LTS release at consistent versions and receive security updates.  This reduces the frequency of failure if users ever re-run docker builds when an external dep has moved to a new version, or been built against newer releases, or is not maintained etc.  

That said, I think we're still working out precise design strategies here around both Python package management (e.g. use of virutalenvs, system python, to conda or not conda, etc -- see #670 , still some unresolved questions)  and our management of OSGeo versions.  @mdsumner I have a [draft branch](https://github.com/rocker-org/rocker-versioned2/blob/patch/osgeo/scripts/install_dev_osgeo.sh) of some further changes to `install_dev_osgeo.sh` that is still a work-in-progress, I'd love your thoughts on that.  

Also tagging @yuvipanda who is on top of the python space and also interested in rocker shipping some newer gdal versions (possibly through the `rocker/binder` pattern?  


To add to complexity, I'm interested in exploring parallels between rocker and pangeo docker images.  Both are Ubuntu-LTS based but take some really different choices, especially around things like GPU support (Rocker builds on Nvidia's official ubuntu-LTS-based images, pangeo/pytorch-notebook does something different, I think via conda).  Of course that's way more than we want to bite off in one PR, but maybe will need some time for brainstorming and planning.

mdsumner:
ah very good, ok treat this PR as merely a wish-list exporation - I'll chase down the bigger picture andideas here, thanks!  

Getting a consistent framework here with Pangeo would be amazing, and I'll also add in the OpenDataCube which is something I especially need side by side R  so I'm pushing at Geoscience Australia to expose their OpenDataCube sandbox docker resources: https://www.opendatacube.org/install  (it was cube-in-a-box but that has fallen aside as far as I can see). 

I'd like to see if that can be streamlined in here and doesn't seem too hard  (I need to explore the other threads here first). Much appreciated!

yuvipanda:
> Getting a consistent framework here with Pangeo would be amazing

I think the primary 3 'big' upstream players here are rocker, pangeo as well as jupyter/docker-stacks. I'm involved in the latter two, and I too would love to see more interop / co-operation between all of these :)

I opened https://github.com/pangeo-data/pangeo-docker-images/issues/163 a while ago to get a R based image into pangeo-docker-stacks, although I now fully believe that should be handled instead by rocker.

PR #714, opened 2023-10-18
binder: Install packages required for Jupyter to PDF convert

yuvipanda:
Per https://github.com/jupyter/nbconvert/issues/1328, these are the packages needed for Jupyter to be able to convert to PDF. Without it, you get unfriendly errors like [this] (https://github.com/2i2c-org/infrastructure/issues/3188#issuecomment-1755969247).

Given that the binder image includes jupyter, and texlive is included in the base, I hope it would be reasonable for the binder image to include enough texlive packages for PDF conversion to work. Otherwise, it works in RStudio but not in Jupyter.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,9 @@
 ### Changes in pre-built images
 
 - `rocker/binder` now includes necessary LaTeX packages to allow Jupyter Notebooks to be converted to PDF. ([#714](https://github.com/rocker-org/rocker-versioned2/pull/714))
+  - `install_jupyter.sh` will now install TexLive if it is not installed, and add
+     additional LaTeX packages needed for Jupyter PDF conversion. This will increase
+     the size of the final image in cases `install_jupyter.sh` is called.
 
 ## 2023-04
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,11 @@
 # News
 
+## 2023-10
+
+### Changes in pre-built images
+
+- `rocker/binder` now includes necessary LaTeX packages to allow Jupyter Notebooks to be converted to PDF. ([#714](https://github.com/rocker-org/rocker-versioned2/pull/714))
+
 ## 2023-04
 
 ### Changes in pre-built images
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,9 +5,12 @@
 ### Changes in pre-built images
 
 - `rocker/binder` now includes necessary LaTeX packages to allow Jupyter Notebooks to be converted to PDF. ([#714](https://github.com/rocker-org/rocker-versioned2/pull/714))
-  - `install_jupyter.sh` will now install TexLive if it is not installed, and add
-     additional LaTeX packages needed for Jupyter PDF conversion. This will increase
-     the build time and size of the final image in cases `install_jupyter.sh` is called.
+
+### Changes in rocker_scripts
+
+- `install_jupyter.sh` will now install TexLive if it is not installed, and add
+  additional LaTeX packages needed for Jupyter PDF conversion. This will increase
+  the build time and size of the final image in cases `install_jupyter.sh` is called.
 
 ## 2023-04
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,7 +7,7 @@
 - `rocker/binder` now includes necessary LaTeX packages to allow Jupyter Notebooks to be converted to PDF. ([#714](https://github.com/rocker-org/rocker-versioned2/pull/714))
   - `install_jupyter.sh` will now install TexLive if it is not installed, and add
      additional LaTeX packages needed for Jupyter PDF conversion. This will increase
-     the size of the final image in cases `install_jupyter.sh` is called.
+     the build time and size of the final image in cases `install_jupyter.sh` is called.
 
 ## 2023-04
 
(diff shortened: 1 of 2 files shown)


eitsupi:
> tlmgr isn't found in and the script is failing, but I do see `tlmgr` in PATH when I'm running the container.

I think `tlmgr` is installed via this script.
https://github.com/rocker-org/rocker-versioned2/blob/c6152e1d5db5e29ee41d8bc0873de9b0ef17861b/scripts/install_texlive.sh

If we want to add these packages to `rocker/binder`, we may need to add a new script.
@cboettig Thoughts?

cboettig:
:+1:  Having support for Myst in binder would be nice.  I think it makes sense to to call the `install_texlive.sh` first to add tlmgr to binder.

scripts/install_jupyter.sh
```diff
@@ -36,6 +36,15 @@ install2.r --error --skipmissing --skipinstalled -n "$NCPUS" remotes
...
 R --quiet -e 'IRkernel::installspec(user = FALSE)'
 
+## Install texlive
+/rocker_scripts/install_texlive.sh
+# Install tex packages needed for Jupyter's nbconvert to work correctly & convert to PDF
+# Sourced from https://github.com/jupyter/nbconvert/issues/1328
+tlmgr install adjustbox caption collectbox enumitem environ eurosym etoolbox jknapltx parskip \
+    pdfcol pgf rsfs tcolorbox titling trimspaces ucs ulem upquote \
+    ltxcmds infwarerr iftex kvoptions kvsetkeys float geometry amsmath fontspec \
+    unicode-math fancyvrb grffile hyperref booktabs soul ec
```
eitsupi, line 46: Could you please make this optional?
If my understanding is correct, the expected behavior of this script is to set jupyter, and this should not be done by default.

And check if `tlmgr` is already installed before exec `install_texlive.sh`.

eitsupi (changes requested):

scripts/install_jupyter.sh
yuvipanda: I'll check for tlmgr before installing!

yuvipanda: I'd consider 'does export to PDF from jupyter work?' to be part of 'set jupyter'. In the jupyter community we have found it important enough that it's part of the minimal notebook package in the jupyter equivalent of rocker (https://github.com/jupyter/docker-stacks/blob/main/docs/using/selecting.md#jupyterminimal-notebook). In the same way I think RStudio users would expect to knit, I think Jupyter users expect to be able to print to PDF. So I'd strongly hope this can be kept for the default installation, as there's only one rocker image that has Jupyter in it.

yuvipanda: I've made the install conditional now

eitsupi: Thanks for looking into that.
I haven't personally used this, so I'll leave the final judgment to @cboettig.
(However, I personally would like to see this become an opt-in feature.)

Also, since this is a big change, could you please update the NEWS?

cboettig: I think this makes sense as the default given the expectations from the Jupyter community.  Definitely agree with mentioning this in NEWS.  Otherwise looks good to me!

yuvipanda: Thanks @cboettig and @eitsupi. I've added an entry to the NEWS file.

NEWS.md
```diff
@@ -1,5 +1,11 @@
 # News
 
+## 2023-10
+
+### Changes in pre-built images
+
+- `rocker/binder` now includes necessary LaTeX packages to allow Jupyter Notebooks to be converted to PDF. ([#714](https://github.com/rocker-org/rocker-versioned2/pull/714))
+
```
eitsupi, line 8: This is also a script change, so please write about that too.
(Since it is not possible to revert to the original behavior, I think it is necessary to emphasize the increased execution time and the size increasing of the image)

eitsupi (changes requested):

NEWS.md
yuvipanda: @eitsupi how about now?

NEWS.md
```diff
@@ -1,5 +1,14 @@
 # News
 
+## 2023-10
+
+### Changes in pre-built images
+
+- `rocker/binder` now includes necessary LaTeX packages to allow Jupyter Notebooks to be converted to PDF. ([#714](https://github.com/rocker-org/rocker-versioned2/pull/714))
+  - `install_jupyter.sh` will now install TexLive if it is not installed, and add
+     additional LaTeX packages needed for Jupyter PDF conversion. This will increase
+     the build time and size of the final image in cases `install_jupyter.sh` is called.
```
eitsupi, line 10:
```suggestion
- `rocker/binder` now includes necessary LaTeX packages to allow Jupyter Notebooks to be converted to PDF. ([#714](https://github.com/rocker-org/rocker-versioned2/pull/714))

### Changes in rocker_scripts

- `install_jupyter.sh` will now install TexLive if it is not installed, and add
  additional LaTeX packages needed for Jupyter PDF conversion. This will increase
  the build time and size of the final image in cases `install_jupyter.sh` is called.
```

NEWS.md
eitsupi: Thanks!

yuvipanda:
Yay, thanks @eitsupi.

The julia test failure looks intermittent and unrelated.

eitsupi (approved):
Ok, merging now.

yuvipanda:
Yay! Thank you, @eitsupi! I appreciate it :)

Would this trigger a rebuild of the various rocker/binder: images? If not, would it be possible to kick off a rebuild of those?

eitsupi:
> Would this trigger a rebuild of the various rocker/binder: images? If not, would it be possible to kick off a rebuild of those?

Today's new release of RStudio Server has already triggered a build.
https://github.com/rocker-org/rocker-versioned2/actions/runs/6587176060

PR #718, opened 2023-11-02
Install python into a venv

yuvipanda:
- Continues to use python and venv from Ubuntu LTS repositories, so they are supported as with everything else that is gotten from apt (see https://github.com/rocker-org/rocker-versioned2/issues/670#issuecomment-1619748511)
- Doesn't currently change any permissions, so present behavior is preserved. However, in the future, we should probably change ownership so end users can install packages in there at runtime (see https://github.com/rocker-org/rocker-versioned2/issues/670#issuecomment-1624691981)
- Sets the VIRTUAL_ENV environment variable to path of the venv we create. This is what the `source activate` script does, and Reticulate also looks for this to discover which python to use (see point 4 of https://rstudio.github.io/reticulate/articles/versions.html#order-of-discovery)
- Sets up PATH appropriately, so python and python3 refer to what is in our venv. This, along with the previous step, ensures same behavior as users typing `source ${VIRTUAL_ENV}/bin/activate` without actually having to do that, preserving end user behavioral semantics. See https://github.com/rocker-org/rocker-versioned2/issues/670#issuecomment-1601860904
- RStudio is also told about new `PATH` and `VIRTUAL_ENV`, using the same pattern as `install_texlive.sh`
- Remove the explicit symlink of python3 -> python, as venv handles this automatically.
- `install_python.sh` now needs to be `source`d, following same pattern as `install_texlive.sh`

Decisions to be made:

- Where do we set the appropriate env variables (VIRTUAL_ENV and PATH)? They need to be set for `install_python.sh` to work correctly. I've set them in the binder image for now, but it should probably be set on a more base image. This is a no-op if `install-python.sh` is not called anywhere.

TODO:
- [x] Update `NEWS` (can be done once everything else is finalized)

Ref https://github.com/rocker-org/rocker-versioned2/issues/670

### DIFF
--- a/dockerfiles/binder_devel.Dockerfile
+++ b/dockerfiles/binder_devel.Dockerfile
@@ -7,7 +7,7 @@ LABEL org.opencontainers.image.licenses="GPL-2.0-or-later" \
 
 ENV NB_USER=rstudio
 ENV VIRTUAL_ENV=/opt/venv
-ENV PATH=${PATH}:${VIRTUAL_ENV}/bin
+ENV PATH=${VIRTUAL_ENV}/bin:${PATH}
 
 RUN /rocker_scripts/install_jupyter.sh
 
--- a/dockerfiles/binder_devel.Dockerfile
+++ b/dockerfiles/binder_devel.Dockerfile
@@ -6,6 +6,8 @@ LABEL org.opencontainers.image.licenses="GPL-2.0-or-later" \
       org.opencontainers.image.authors="Carl Boettiger <cboettig@ropensci.org>"
 
 ENV NB_USER=rstudio
+ENV VIRTUAL_ENV=/opt/venv
+ENV PATH=${PATH}:${VIRTUAL_ENV}/bin
 
 RUN /rocker_scripts/install_jupyter.sh
 
--- a/scripts/install_python.sh
+++ b/scripts/install_python.sh
@@ -40,7 +40,8 @@ python3 -m pip --no-cache-dir install --upgrade \
 
 # Make the venv owned by the staff group, so users can install packages
 # without having to be root
-chown -R rstudio:staff /opt/venv
+chown -R root:staff "${VIRTUAL_ENV}"
+chmod g+ws "${VIRTUAL_ENV}"
 
 install2.r --error --skipmissing --skipinstalled -n "$NCPUS" reticulate
 
--- a/scripts/install_python.sh
+++ b/scripts/install_python.sh
@@ -22,13 +22,13 @@ apt_install \
     python3-venv \
     swig
 
+# Setup a virtualenv to install things into
+python3 -m venv ${VIRTUAL_ENV}
+export PATH=${VIRTUAL_ENV}:${PATH}
+
+# Upgrade version of pip inside the virtualenv
 python3 -m pip --no-cache-dir install --upgrade \
     pip
-
-# Some TF tools expect a "python" binary
-if [ ! -e /usr/local/bin/python ]; then
-    ln -s "$(which python3)" /usr/local/bin/python
-fi
 
 install2.r --error --skipmissing --skipinstalled -n "$NCPUS" reticulate
 
--- a/scripts/install_python.sh
+++ b/scripts/install_python.sh
@@ -61,9 +61,10 @@ if [ "${UBUNTU_CODENAME}" == "focal" ]; then
 fi
 
 # Check that python and python3 point to correct places
-echo "Check python and python3 executables point to the correct place..."
+echo "Check python, python3 and pip executables point to the correct place..."
 echo "python -> $(which python)"
 echo "python3 -> $(which python3)"
+echo "pip -> $(which pip)"
 
 # Check Python version
 echo -e "Check the Python to use with reticulate...\n"
--- a/scripts/install_python.sh
+++ b/scripts/install_python.sh
@@ -38,6 +38,10 @@ python3 -m venv "${VIRTUAL_ENV}"
 python3 -m pip --no-cache-dir install --upgrade \
     pip
 
+# Make the venv owned by the staff group, so users can install packages
+# without having to be root
+chown -R rstudio:staff /opt/venv
+
 install2.r --error --skipmissing --skipinstalled -n "$NCPUS" reticulate
 
 # Clean up
--- a/scripts/install_python.sh
+++ b/scripts/install_python.sh
@@ -56,6 +56,11 @@ if [ "${UBUNTU_CODENAME}" == "focal" ]; then
     fi
 fi
 
+# Check that python and python3 point to correct places
+echo -e "Check python and python3 executables point to the correct place...\n"
+which python
+which python3
+
 # Check Python version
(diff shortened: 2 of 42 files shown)


yuvipanda:
Created this draft mostly to see what CI says!

yuvipanda:
ok, this is ready for a first round of review :)

dockerfiles/binder_devel.Dockerfile
```diff
@@ -6,6 +6,8 @@ LABEL org.opencontainers.image.licenses="GPL-2.0-or-later" \
       org.opencontainers.image.authors="Carl Boettiger <cboettig@ropensci.org>"
 
 ENV NB_USER=rstudio
+ENV VIRTUAL_ENV=/opt/venv
```
cboettig, line 9: nice, will test to confirm but I think reticulate should find venvs in `/opt`

scripts/install_python.sh
```diff
@@ -22,14 +22,22 @@ apt_install \
...
     swig
 
+# Setup a virtualenv to install things into
+
+# Put things under /opt/venv, if nothing else is specified
+export VIRTUAL_ENV="${VIRTUAL_ENV:=/opt/venv}"
+export PATH="${VIRTUAL_ENV}/bin:${PATH}"
+
+# Make sure that Rstudio sees these env vars too
+echo "PATH=${PATH}" >>"${R_HOME}/etc/Renviron.site"
```
cboettig, line 32: @eitsupi did we have some logic elsewhere to avoid repeatedly writing new definitions of PATH to `Renviron.site` ?  It's probably fine but if this script is run repeatedly it can make that file a bit clunky, and I thought I recalled (or imagined?) you having some nice trick to keep it tidy?

stacks/devel.json
```diff
@@ -127,7 +127,9 @@
       },
       "FROM": "rocker/geospatial:devel",
       "ENV": {
-        "NB_USER": "rstudio"
+        "NB_USER": "rstudio",
+        "VIRTUAL_ENV": "/opt/venv",
+        "PATH": "${PATH}:${VIRTUAL_ENV}/bin"
       },
```
cboettig, line 133: This would only sets these new envs on the devel image (i.e. image based on R-devel) of geospatial.  I think we really want to set these in any of the images that call `install_python.sh` (or `install_jupyter.sh`) in their RUN block, on both devel and 4.3.2 (aka latest) stack files, right? i.e. on https://github.com/rocker-org/rocker-versioned2/blob/master/stacks/4.3.2.json, for every stack entry that includes a call to `install_python.sh`, right?  (apologies @eitsupi if I've miss-remembered where these definitions get set in the current build process!)

cboettig (changes requested):

stacks/devel.json
yuvipanda: I agree I'd like it to be on all the jsons :) I had initially thought that it would need to be here and the other JSON stuff is autogenerated off this by make (hence me opening #720), but I am probably wrong? :D Should I just manually add it to all the json files?

eitsupi: When we run this script, the two latest stack files will be updated. For older versions, please update them manually.

https://github.com/rocker-org/rocker-versioned2/blob/56ce3b840cf7f79a809676d09372a71519f21c47/build/make-stacks.R#L503

eitsupi:
What is the current status of this?

yuvipanda:
@cboettig gave me some feedback via slack that I need to dig into. @cboettig can you put them here too?

cboettig:
Thanks @yuvipanda !  

Yup, I think the main thing is considering changing the permissions on `/opt/venv` from root to something else.  (In the R site library we put these in the `staff` group, allowing any user added to that group to update or add packages to the shared site library.  Not sure how nicely pip plays with group-based vs user-based permissions though?  In my use/testing of this image I've just been making `/opt/venv` owned by the default user, https://github.com/boettiger-lab/nasa-topst-env-justice/blob/main/.devcontainer/Dockerfile#L16

A second more minor issue is that under this setup, the default `python` now points to `/opt/venv/bin/python` as intended, but if a user calls `pip install` instead of `python -m pip install`, they still get `/usr/bin/pip` instead of `/opt/venv/bin/pip`, which creates a gotcha that it would be better to avoid if possible??

yuvipanda:
> A second more minor issue is that under this setup, the default python now points to /opt/venv/bin/python as intended, but if a user calls pip install instead of python -m pip install, they still get /usr/bin/pip instead of /opt/venv/bin/pip, which creates a gotcha that it would be better to avoid if possible??

Hmm, since there should be a `pip` in `/opt/venv/bin/pip`, if `/opt/venv/bin` is in $PATH, that should mean `pip` should call that pip by default, right? Is that not what's happening?

yuvipanda:
Making `/opt/venv` be owned by `staff` seems reasonable to me. If it works for R, my intuition is that it will also work for Python!

cboettig:
@eitsupi ok just tested and this looks ready to merge to me!

cboettig (approved):

yuvipanda:
Yay that's awesome, thank you @cboettig!

scripts/install_python.sh
```diff
@@ -22,13 +22,25 @@ apt_install \
...
 python3 -m pip --no-cache-dir install --upgrade \
     pip
 
-# Some TF tools expect a "python" binary
-if [ ! -e /usr/local/bin/python ]; then
-    ln -s "$(which python3)" /usr/local/bin/python
-fi
+# Make the venv owned by the staff group, so users can install packages
+# without having to be root
+chown -R rstudio:staff /opt/venv
```
eitsupi, line 43: Don't hard-code the username. There is no guarantee that the user exists.

eitsupi (changes requested):

scripts/install_python.sh
yuvipanda: @eitsupi in `install_texlive.sh` I see:

```
NON_ROOT_USER=$(getent passwd "1000" | cut -d: -f1)
```

and this is used as username. Should I do that?

eitsupi: Is there a bad reason not to be the root user?
If so, please do so.

However, in that case, I don't think it will work if the uid is anything other than 1000

yuvipanda: Ok, given I was trying to replicate what happens in R, I've set this to match what is done to `$R_HOME/site-library`

eitsupi (approved):
The tests seem passed. Thanks!

eitsupi:
Could you update the NEWS file?

yuvipanda:
I always seem to struggle with updating NEWS but gave it a shot!

eitsupi:
Looks good.

The CI failure is notthing to do with changes in this PR, so merging.

yuvipanda:
Awesome! Thanks a lot, @eitsupi! Will this trigger a rebuild of the images?

eitsupi:
Sorry, my review was not thorough enough (i.e. this should not have been merged yet), the fix for the stack file is completely incomplete here.
We had to update it manually for all versions.
See the issue #739.

PR #722, opened 2023-11-03
[CI] Fix and refactor the script to auto update definition files

eitsupi:
- Fix a process that started to fail due to the announcement of Ubuntu 24.04.
- Use short URLs `p3m.dev/cran` for Posit Public Package Manager, which have recently come into use.


eitsupi:
Super Linter failure on CI is unrelated to this change.

eitsupi:
@cboettig Could you check this?

cboettig:
nice this is working for me!

eitsupi:
Thanks. Merging.

PR #721, opened 2023-11-02
Update tex packages needed for knitting Rmd files to PDF

yuvipanda:
If I create a new Rmarkdown file and try to knit to PDF, it has to install a big bunch of packages (which sometimes seems to take minutes, at least on my laptop) before it can actually knit. And this seems to happen each time the container starts.

My guess is that the list of tex packages needed for rmd -> pdf knitting to work has just increased over time.

I've added all the packages that were manually installed, by just looking at the 'render' tab in RStudio as installation was happening, and adding them. They are sorted alphabetically.

I've removed any packages here that were also present in `install_jupyter.sh` to avoid duplication.

<details>
<summary>Output from render tab indicating the packages that were being installed</summary>
<pre>
processing file: t.Rmd

output file: t.knit.md

running mktexlsr ...
done running mktexlsr.
tlmgr: package log updated: /opt/texlive/texmf-var/web2c/tlmgr.log
tlmgr: command log updated: /opt/texlive/texmf-var/web2c/tlmgr-commands.log
tlmgr update --all --self
tlmgr: package repository https://www.texlive.info/tlnet-archive/2023/10/30/tlnet (not verified: gpg unavailable)
tlmgr install: package already present: texlive-scripts
[1/1, ??:??/??:??] install: ec [258k]
running mktexlsr ...
done running mktexlsr.
tlmgr: package log updated: /opt/texlive/texmf-var/web2c/tlmgr.log
tlmgr: command log updated: /opt/texlive/texmf-var/web2c/tlmgr-commands.log
tlmgr: package repository https://www.texlive.info/tlnet-archive/2023/10/30/tlnet (not verified: gpg unavailable)
[1/1, ??:??/??:??] install: kvoptions [8k]
running mktexlsr ...
done running mktexlsr.
tlmgr: package log updated: /opt/texlive/texmf-var/web2c/tlmgr.log
tlmgr: command log updated: /opt/texlive/texmf-var/web2c/tlmgr-commands.log
tlmgr: package repository https://www.texlive.info/tlnet-archive/2023/10/30/tlnet (not verified: gpg unavailable)
[1/1, ??:??/??:??] install: ltxcmds [5k]
running mktexlsr ...
done running mktexlsr.
tlmgr: package log updated: /opt/texlive/texmf-var/web2c/tlmgr.log
tlmgr: command log updated: /opt/texlive/texmf-var/web2c/tlmgr-commands.log
[...]

PR #741, opened 2024-01-02
Add tlmgr packages needed for quarto to knit to PDF

yuvipanda:
Follow-up to https://github.com/rocker-org/rocker-versioned2/pull/721, adding tlmgr packages that *quarto* seems to need to knit to PDF. Without these, it takes a minute or more (depending on network) for first knit to happen apparently.

Reported to me by users from University of Toronto

### DIFF
--- a/scripts/install_texlive.sh
+++ b/scripts/install_texlive.sh
@@ -56,6 +56,7 @@ tlmgr install \
     bibtex \
     bigintcalc \
     bitset \
+    bookmark \
     context \
     ec \
     epstopdf-pkg \
@@ -70,6 +71,7 @@ tlmgr install \
     inconsolata \
     infwarerr \
     intcalc \
+    koma-script \
     kvdefinekeys \
     kvoptions \
     kvsetkeys \
@@ -77,6 +79,7 @@ tlmgr install \
     listings \
     ltxcmds \
     makeindex \
+    mdwtools \
     metafont \
     mfware \
     parskip \
@@ -87,6 +90,7 @@ tlmgr install \
     rerunfilecheck \
     stringenc \
     tex \
+    tikzfill \
     tools \
     uniquecounter \
     url \


cboettig:
Thanks @yuvipanda , this looks good to me.

eitsupi:
Given that some users don't need these packages and it's not clear what packages Quarto requires, I doubt whether it's worth adding them.

I think it depends on the purpose of this script.
Does it really matter that it takes 1 minute?

eddelbuettel:
The difficulty with latex 'on demand' issues is that they depend on the document in question.  The only real answer, if someone really needs latex, maybe adding texlive.  Else '1 min' once per run is not so bad, really.

yuvipanda:
For me, the goal is to make sure that the *simplest possible document* will knit to PDF. I think of those as 'these are the requirements for quarto' - similar to software package requirements, just in a different ecosystem. The packages added in this PR were required for the hello world of quarto document to knit:

<details>
<summary>Simple document that was tried to knit</summary>
<pre>
---
title: "Untitled"
format: pdf
editor: visual
---

## Quarto

Quarto enables you to weave together content and executable code into a finished document. To learn more about Quarto see <https://quarto.org>.

## Running Code

When you click the **Render** button a document will be generated that includes both content and the output of embedded code. You can embed code like this:

```{r}
1 + 1
```

You can add options to executable code like this

```{r}
#| echo: false
2 * 2
```

The `echo: false` option disables the printing of code (only output is displayed).

$$
y=X\beta+\epsilon.
$$

```{r}
rnorm(10)
```
</pre>
</details>

So while additional tex packages that may be needed due to using extra features can be installed on demand, I think it's helpful to install the packages included in this PR as a baseline for 'support quarto'.

yuvipanda:
The sample document I showed is a slight modification of what RStudio opens if you go to `File -> New File -> Quarto Document`

cboettig:
Given that we already include some tlmgr packages, I think it makes sense that the 'out-of-the-box' dependencies on tlmgr for features already included in this docker image (i.e. rocker/verse) include all those which would be needed for default behaviors: 

- building an R package (with pdf documentation)
- building a quarto PDF document using the default quarto pdf format.


I agree with @eitsupi that some users don't needs these packages.  but that is also why we add latex late in the 'stack' (at `verse`, not in `rocker/r-ver`, `rstudio`, or `tidyverse`, and why we don't install (the several extra gigabyte) texlive dependencies. I agree with @eddelbuettel that 1 min isn't so bad in general, but I think default templates should be built faster, and it just feels a bit sloppy right now that we have missed some of those packages.  I don't think we want the list to creep up (e.g. to include all deps of, say, `rticles` tlmgr dependencies, or reinvent the entire texlive package groups), but I think the set @yuvipanda has identified here is sensible.  This PR would add very little to the size of the base images, have little chance of otherwise breaking downstream users, and strengthen this image in a clearly defined and limited-scope way to be 'self-consistent' with the out-of-the-box behavior.  (Yes, future versions of quarto / pandoc pdf templates could lead to future changes here, but I suspect overall that scope is pretty limited).  

So anyway, I appreciate all the points in general here but I'm in favor of merging this particular set.

eitsupi:
Then let's merge this. The increase in capacity seems to be about 20MB.

yuvipanda:
Thank you, @cboettig @eddelbuettel and @eitsupi for your thoughtful stewardship of this project :)

Issue #736, opened 2023-12-08
Nvidia devel tools get removed from rocker/cuda-based images


javdg:
### Container image name

rocker/cuda

### What operating system related to this question?

Linux

### Question

As they are based on nvidia/cuda:11.8.0-cudnn8-devel-ubuntu22.04 ([dockerfiles/cuda_4.3.0.Dockerfile](https://github.com/rocker-org/rocker-versioned2/blob/de8b815b1b23c368308cc9dc960cb8a7c724be9f/dockerfiles/cuda_4.3.0.Dockerfile)), I was expecting to have the full CUDA Toolkit devel environment available when using any of the rocker/cuda and derived images.  
However, https://github.com/rocker-org/rocker-versioned2/blob/de8b815b1b23c368308cc9dc960cb8a7c724be9f/scripts/install_R_source.sh#L160C39 seems to remove relevant packages (although not explicitly specified). This includes `cuda-compiler-11-8*, cuda-minimal-build-11-8*, cuda-nvcc-11-8*` and marks others for subsequent autoremoval (`cuda-cuxxfilt-11-8, cuda-nvprune-11-8`).  
This does e.g. prevent the torch package to be easily installed with GPU-support within such containers.  
Is this done intentionally?

eitsupi:
@cboettig Any thoughts?

benz0li:
Interesting. This does not happen with my CUDA-enabled JupyterLab docker stacks which are also based on `nvidia/cuda:11.8.0-cudnn8-devel-ubuntu22.04`.

I use

```bash
apt-get -y purge PACKAGES
apt-get -y autoremove
rm -rf /var/lib/apt/lists/*
```
 
instead of

```bash
apt-get remove --purge -y PACKAGES
apt-get autoremove -y
apt-get autoclean -y
rm -rf /var/lib/apt/lists/*
```

eitsupi:
@cboettig Could you take a look at this?

cboettig:
@eitsupi thanks for the ping, yeah I'll take a look!

cboettig:
Yup, 

>  somehow apt-get remove --purge -y ${BUILDDEPS}; apt-get autoremove -y removes relevant CUDA packages from the Rocker images.

it looks like some of the build deps are (unsurprisingly) also build deps of cuda-devel.  Still seems a bit puzzling to me that it would grab some of the nvidia tools. 

[...]

PR #740, opened 2023-12-27
Manually set PATH and VIRTUAL_ENV for all binder related files

yuvipanda:
Ref https://github.com/rocker-org/rocker-versioned2/issues/739

PR #743, opened 2024-01-07
Patch cuda dev tools (#736)

cboettig:
Just a draft proposal here so far, but I think the least-friction way to do this.  Doesn't impact the normal r-ver stack. In the cuda stack, the idea would be to copy the compiled version of R from r-ver rather than run the typical `install_R_source.sh` script, which creates the problems in purging dependencies.  To make this work the runtime dependencies still need to be installed, so for simplicity I've added these in another script.  (technically that means we have these dependencies specified in two places now, so I could adjust `install_R_source.sh` to source the new script instead, not sure if that's preferable?). 

I haven't taken a go at defining the new bakefile template yet.  Hopefully straight forward, though will need to make sure it gets the right version tag for r-ver images in the non-latest images; I haven't thought out the logic there.  

@eitsupi thoughts?


eitsupi:
Thanks for working on this!
I think this is a good direction, but looking at the original script, wouldn't it be cleaner and easier to add an option to skip the part that is building R?

https://github.com/rocker-org/rocker-versioned2/blob/6efce3a3731875f399c5e0ab1397f352c9e145d3/scripts/install_R_source.sh#L98-L139

cboettig:
Thanks @eitsupi !  Yes, great point, that does sound like a much simpler solution!  I'll scratch this and take a go at that instead.

cboettig:
@eitsupi okay looping back to this using your simpler mechanism. Can you take a look over this when you get a chance?  Thanks!

eitsupi:
I'm sorry my comment is confusing.
What I was trying to say was that rather than adding a new script, it would be better to use the same script with the flag added to `install_R` and skipped the "build R" steps so that we don't have to have almost identical copies in two files.

scripts/install_R_source.sh
```diff
@@ -157,7 +158,9 @@ rm -rf "R.tar.gz"
 cp /usr/bin/checkbashisms /usr/local/bin/checkbashisms
 
 # shellcheck disable=SC2086
-apt-get remove --purge -y ${BUILDDEPS}
+if [ "${PURGE_BUILDDEPS}" == "TRUE" ]; then
```
cboettig, line 161: @eitsupi you said:

> it would be better to use the same script with the flag added to install_R_source.sh and skipped the "build R" steps so that we don't have to have almost identical copies in two files.

I am adding a flag to install_R_source.sh right here though.  what two files are you referring two as almost identical copies?

eitsupi: Perhaps the assumptions were not engaged. I had imagined something like the following:

```dockerfile
FROM rocker/r-ver:4.3.2 as builder

FROM FROM nvidia/cuda:11.8.0-cudnn8-devel-ubuntu22.04
COPY --from=builder /foo/bar
COPY scripts/install_R_source.sh /rocker_scripts/install_R_source.sh
ENV SKIP_BUILD_R="true"
RUN /rocker_scripts/install_R_source.sh
```

cboettig: Thanks, I see what you mean.  That looks rather more complicated though. 
-  It requires using the multi-stage build, and i'm not sure how easy that is to add to our existing build stack.   
- Copying `/foo/bar` isn't so clean, because our standard build installs R_HOME into /usr/local, with binaries in /usr/local/bin, etc, rather than freestanding.  Yes, I proposed a solution around this in my previous proposal, but it is a significant change that could break some user expectations.
- SKIP_BUILD_R would have to wrap around thin install-time deps,  building of R and and clean up and purge of dependencies.  That's fine, but it feels more natural to me to have an option that just avoids the `apt remove` step.  After all, given that `apt remove` has unexpected consequences, it seems reasonable to be able to opt out of that without opting out of the rest of it.  and intuitively, it's easier I think to guess what `install_R_source.sh` does behaves in response to a option of `PURGE_BUILDDEPS=FALSE` than it is to guess what `install_R_source.sh` does in response to an option of `SKIP_BUILD_R`.

What's the downside of the approach here? I still don't see what's been duplicated.

cboettig:
apologies about my lint errors, looks like a whitespace issue but I'm not sure.

Happy to change strategies again if that's better.  I think adding the option of simply turning off the uninstall of builddeps is the least invasive way to address this issue with minimal changes.  Please let me know if there's a a better alternative, but I think I'd rather avoid introducing staged builds merely to address the problems created by the `apt-remove` script.  (If we're going to introduce support for staged builds, I think it makes most sense to do so in a refactor of `r-ver`, where that is the natural mechanism in modern Docker to accomplish what we do via apt remove line, as you know.)

I did slip a few other changes in here that look like they got missed in #740 (that updated the necessary env vars for most but not all recipes that use the the revised install_python.sh script, but I think we want that on cuda images too).  This also corrects a small permissions issue in install_python.sh.  Let me know if it would be better to do these bits as a separate PR, just thought it made sense to fix them while I was testing these cuda images anyhow.  

Thanks for the review and suggestions!

stacks/4.3.2.json
```diff
@@ -258,7 +258,7 @@
         "PYTHON_CONFIGURE_OPTS": "--enable-shared",
         "RETICULATE_AUTOCONFIGURE": "0",
         "PURGE_BUILDDEPS": "FALSE",
-        "VIRUTAL_ENV": "/opt/venv",
+        "VIRTUAL_ENV": "/opt/venv",
```
eddelbuettel, line 261: Hah!

eitsupi:
@cboettig I understood. Indeed, I think this simple solution can be implemented relatively quickly and is worthwhile.

By the way, the Lint error looks like an indent size issue.

scripts/install_R_source.sh
```diff
@@ -11,6 +11,7 @@
 set -e
 
 R_VERSION=${1:-${R_VERSION:-"latest"}}
+PURGE_BUILDDEPS=${PURGE_BUILDDEPS=-"TRUE"}
```
eitsupi, line 14: I think bools in env vars should be `"true"` or `"false"`, not `"TRUE"` or `"FALSE"`.

stacks/devel.json
```diff
@@ -148,15 +148,17 @@
         "org.opencontainers.image.title": "rocker/cuda",
         "org.opencontainers.image.description": "NVIDIA CUDA libraries added to Rocker image."
       },
-      "FROM": "nvidia/cuda:11.8.0-cudnn8-devel-ubuntu24.04",
+      "FROM": "nvidia/cuda:11.8.0-cudnn8-devel-ubuntu22.04",
```
eitsupi, line 151: If I remember correctly, this value is updated by the script, so don't update it here manually.

eitsupi (changes requested):

scripts/install_R_source.sh
eddelbuettel: I _think_ env vars are untyped so 1 and "1" and "TRUE" and "True" and "true" are all the same.  Unless, of course, the receiving end makes a difference when it parses.  Otherwise the check is commonly 'set or not set'.

stacks/devel.json
eddelbuettel: I wondered about this when the script added it.  Are we sure a) the production cycle for Ubuntu 24.04 has started _and_ b) that NVidia already publishes to it?  I was under the impression NVidia was particularly 'laggy' ie the `gcc` version implied by `nvcc` was always ages behind.

benz0li: There is no `nvidia/cuda:11.8.0-cudnn8-devel-ubuntu24.04`. There are

* `nvidia/cuda:11.8.0-cudnn8-devel-ubuntu22.04`
* `nvidia/cuda:11.8.0-cudnn8-devel-ubuntu20.04`
* `nvidia/cuda:11.8.0-cudnn8-devel-ubuntu18.04` (EOL)

See https://gitlab.com/nvidia/container-images/cuda/-/tree/master/dist/11.8.0

eitsupi: What I mean is "not read", but "don't set it manually, because if the script is execute by CI, it will be updated".

The priority here is to use the same Ubuntu version as `rocker/r-ver`. I haven't found an easy way to check if that image actually exists (This is definitely possible using buildx, but I couldn't figure out how to check without logging into the Container Registry.), so no such processing is implemented.
If you try to build it, it will simply fail.

Currently pointing to Ubuntu 24.04 is not as intended, I really wanted it to be 22.04, which is the same as the current `ubuntu:latest`. In other words, this is a bug in the script.
https://github.com/rocker-org/rocker-versioned2/blob/7d088379de5b360408539a62f5ed5548a9527cd6/stacks/devel.json#L26

https://github.com/rocker-org/rocker-versioned2/blob/7d088379de5b360408539a62f5ed5548a9527cd6/build/make-stacks.R#L449-L450

scripts/install_R_source.sh
eitsupi: Is that a common practice in shell script?
I saw in bash:

```bash
$ if "true"; then echo "true"; fi
true

$ if "false"; then echo "true"; fi

$ if "1"; then echo "true"; fi
bash: 1: command not found

$ if "TRUE"; then echo "true"; fi
bash: TRUE: command not found

$ if "True"; then echo "true"; fi
bash: True: command not found
```

stacks/devel.json
cboettig: right, so it now reads 22.04, and I think should work as intended.  let me know if anything needs to be changed?

scripts/install_R_source.sh
cboettig: right, but as you I did this a string comparison since booleans in bash always confuse me.  Anyway I've changed these to lower-case now.  Thanks for catching this!  let me know if you'd recommend further changes here.

stacks/devel.json
eitsupi: > right, so it now reads 22.04, and I think should work as intended. let me know if anything needs to be changed?

Run the scripts to update the files. It should be back.

https://github.com/rocker-org/rocker-versioned2/blob/7d088379de5b360408539a62f5ed5548a9527cd6/build/README.md?plain=1#L5-L12

cboettig: @eitsupi ok I see what you mean, `make-stacks.R` tries to figure this out automatically and it's coming up with the nonsense answer of 24.04 as the answer on the devel image.  I still think that's confusing (doesn't that mean we're just failing to build a devel flavor for the whole cuda stack?) but I agree it's a separate issue and not one that needs to be solved in this PR.

I've just pushed an edit to roll that back.  

I thought I fixed whitespace but am still struggling with the linter :-(

cboettig:
@eitsupi any further thoughts or edits you'd like to see here before we merge?

eitsupi:
Don't you need to edit other stack files?

Don't forget to also run other build scripts.

The linter error indicates that the number of indentations is different. (The indent size in other places is 4, but the line you added is 2)

cboettig:
Thanks @eitsupi !  I've fixed the whitespace issue with linting now and run the remaining scripts.

I'd like to at least start with these changes only in the active images (devel and latest, though looks like the scripts automatically back-port this to the most recent locked release as well).  Traditionally we have allowed images to continue to evolve at the front end while trying to keep old versions as stable and locked as possible, at least while still working out features.  In some cases (maybe including this one) we might want to port this to all 4.x releases, but maybe we can revisit that once this is merged and working?

- do we want to run the build scripts manually in the commit as well?  would it make sense to have that task performed by CI?
- if running these four build scripts is our intended workflow for PRs, (or even for testing PRs), I wonder if it would not make sense to have them bundled under a Makefile directive for that?
- we've moved to a more portable linter, right?  Might be nice to have a `make lint` option.

eitsupi:
> * do we want to run the build scripts manually in the commit as well?  would it make sense to have that task performed by CI?
> * if running these four build scripts is our intended workflow for PRs, (or even for testing PRs), I wonder if it would not make sense to have them bundled under a Makefile directive for that?

As your are aware, these scripts run automatically once a day on the main branch, so in many cases there is no problem even if no changes are made on the PR.
However, I was concerned that a merge without recognizing this might cause unexpected changes.

eitsupi (approved):
> I'd like to at least start with these changes only in the active images (devel and latest, though looks like the scripts automatically back-port this to the most recent locked release as well). Traditionally we have allowed images to continue to evolve at the front end while trying to keep old versions as stable and locked as possible, at least while still working out features. In some cases (maybe including this one) we might want to port this to all 4.x releases, but maybe we can revisit that once this is merged and working?

Makes sense. Let's try merging.

eitsupi:
I don't know what the CI failures indicate - TeX related stuff breaks a lot, so I don't know if this is temporary or not. I don't think it has anything to do with this PR.

cboettig:
the tex error confuses me too. Like you say, it's always a fragile point.  locally that test has no problems for me.

scripts/install_R_source.sh
```diff
@@ -11,6 +11,7 @@
 set -e
 
 R_VERSION=${1:-${R_VERSION:-"latest"}}
+PURGE_BUILDDEPS=${PURGE_BUILDDEPS=-"true"}
```
eitsupi, line 14: Maybe typo?
`PURGE_BUILDDEPS=${PURGE_BUILDDEPS:-"true"}`

eitsupi:
See #750 and #751.

This PR breaks a lot of things...

scripts/setup_R.sh
```diff
@@ -68,7 +69,9 @@ if [ ! -x "$(command -v r)" ]; then
 
     # Clean up
     # shellcheck disable=SC2086
-    apt-get remove --purge -y ${BUILDDEPS}
+    if [ "${PURGE_BUILDDEPS}" == "true" ]; then
```
eitsupi, line 72: I think this should be opt-out, not opt-in.
@cboettig  Could you please update this to `if [ "${PURGE_BUILDDEPS}" != "false" ]; then`

eitsupi: Done in #753

PR #750, opened 2024-01-24
install texlive conditionally

cboettig:
@eitsupi my apologies, I didn't realize my local test was pulling the upstream image.  The tex issue is specific to the changes in #743 -- specifically, the official ubuntu repo texlive binaries are installed and purged in the BUILDDEPS list.  By not purging builddeps, we had conflicting versions of texlive.  

This PR modifies the `install_texlive.sh` to not do the manual install of tlmgr from CPAN if the texlive has already be installed by apt-get (i.e. `/usr/bin/latex` already exists).  I believe it should resolve the issue.

### DIFF
--- a/scripts/install_texlive.sh
+++ b/scripts/install_texlive.sh
@@ -1,4 +1,10 @@
 #!/bin/bash
+
+if [[ -x "/usr/bin/latex" ]]; then
+    echo "texlive already installed"
+    exit 0
+fi
+
 set -e
 
 CTAN_REPO=${1:-${CTAN_REPO:-"https://mirror.ctan.org/systems/texlive/tlnet"}}
--- a/scripts/install_texlive.sh
+++ b/scripts/install_texlive.sh
@@ -1,11 +1,10 @@
 #!/bin/bash
+set -e
 
 if [[ -x "/usr/bin/latex" ]]; then
     echo "texlive already installed"
     exit 0
 fi
-
-set -e
 
 CTAN_REPO=${1:-${CTAN_REPO:-"https://mirror.ctan.org/systems/texlive/tlnet"}}
 
--- a/scripts/install_verse.sh
+++ b/scripts/install_verse.sh
@@ -74,9 +74,13 @@ apt-get remove -y systemd
 apt-get -y autoremove
 
 ## Add LaTeX, rticles and bookdown support
-wget "https://travis-bin.yihui.name/texlive-local.deb"
-dpkg -i texlive-local.deb
-rm texlive-local.deb
+
+## tinytex recommends a dummy texlive if using tlmgr manually
+if [[ ! -x "/usr/bin/latex" ]]; then
+    wget "https://travis-bin.yihui.name/texlive-local.deb"
+    dpkg -i texlive-local.deb
+    rm texlive-local.deb
+fi
 
 ## Install texlive
 /rocker_scripts/install_texlive.sh
--- a/scripts/install_verse.sh
+++ b/scripts/install_verse.sh
@@ -74,7 +74,6 @@ apt-get remove -y systemd
 apt-get -y autoremove
 
 ## Add LaTeX, rticles and bookdown support
-
 ## tinytex recommends a dummy texlive if using tlmgr manually
 if [[ ! -x "/usr/bin/latex" ]]; then
     wget "https://travis-bin.yihui.name/texlive-local.deb"
--- a/tests/rocker_scripts/matrix.json
+++ b/tests/rocker_scripts/matrix.json
@@ -17,8 +17,7 @@
     "install_python.sh",
     "install_jupyter.sh",
     "install_julia.sh",
-    "install_cuda-10.1.sh",
-    "install_texlive.sh"
+    "install_cuda-10.1.sh"
   ],
   "script_arg": [
     "none"
@@ -69,7 +68,7 @@
     {
       "base_image": "rocker/cuda",
       "tag": "latest",
-      "script_name": "install_texlive.sh",
+      "script_name": "install_verse.sh",
       "script_arg": "none"
     },
     {
--- a/tests/rocker_scripts/matrix.json
+++ b/tests/rocker_scripts/matrix.json
@@ -68,7 +68,7 @@
     },
     {
       "base_image": "rocker/cuda",
-      "tag": "cuda11.1",
+      "tag": "latest",
       "script_name": "install_texlive.sh",
       "script_arg": "none"
     },
--- a/tests/rocker_scripts/matrix.json
+++ b/tests/rocker_scripts/matrix.json
@@ -17,7 +17,8 @@
     "install_python.sh",
     "install_jupyter.sh",
     "install_julia.sh",
-    "install_cuda-10.1.sh"
+    "install_cuda-10.1.sh",
+    "install_texlive.sh"
   ],
   "script_arg": [
     "none"
@@ -63,6 +64,12 @@
       "base_image": "rocker/r-ver",
       "tag": "4.0.4",
       "script_name": "install_tensorflow.sh",
+      "script_arg": "none"
+    },
+    {
+      "base_image": "rocker/cuda",
+      "tag": "cuda11.1",
+      "script_name": "install_texlive.sh",
       "script_arg": "none"
     },
     {
--- a/scripts/install_jupyter.sh
+++ b/scripts/install_jupyter.sh
@@ -43,12 +43,15 @@ if ! command -v tlmgr; then
     source /rocker_scripts/install_texlive.sh
 fi
 
-# Install tex packages needed for Jupyter's nbconvert to work correctly & convert to PDF
-# Sourced from https://github.com/jupyter/nbconvert/issues/1328
-tlmgr install adjustbox caption collectbox enumitem environ eurosym etoolbox jknapltx parskip \
-    pdfcol pgf rsfs tcolorbox titling trimspaces ucs ulem upquote \
-    ltxcmds infwarerr iftex kvoptions kvsetkeys float geometry amsmath fontspec \
-    unicode-math fancyvrb grffile hyperref booktabs soul ec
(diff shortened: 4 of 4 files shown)


eitsupi:
See https://github.com/rocker-org/rocker-versioned2/pull/743#discussion_r1464155435

I apologize for the poor quality of the review.
I think the last build didn't remove the build dependencies, which increased the image size of `rocker/r-ver` by more than 2GB. It needs to be fixed immediately.

eitsupi:
See #751

scripts/install_texlive.sh
```diff
@@ -1,4 +1,10 @@
 #!/bin/bash
+
+if [[ -x "/usr/bin/latex" ]]; then
+    echo "texlive already installed"
+    exit 0
```
eitsupi, line 5: Why skip all other processes? In that case, the package installation will not take place, right?

eitsupi:
It's really unfortunate that the regression test was working correctly and we ended up merging.

eitsupi:
> I didn't realize my local test was pulling the upstream image.

I don't think the current structure is really good. Ideally, we should only rely on all images `ubuntu` and not on the rocker image that has been pushed.
But I don't have the energy to do it because it takes a lot of effort to change the structure.

scripts/install_texlive.sh
cboettig: I don't believe any of the processes listed inside `install_texlive.sh` ought to take place if we're using the ubuntu binaries which already supply these.

cboettig:
I think the failing test will succeed once https://github.com/rocker-org/rocker-versioned2/actions/runs/7634476675/job/20798413557 is finished.  I'll leave this as a draft until then. 


The current test structure seems to depend on testing against the current published version of the `rocker/r-ver` image, not the modified one, which is what confused me? In https://github.com/rocker-org/rocker-versioned2/actions/runs/7634519606/job/20799340714?pr=750#step:5:31, it appears to be pulling rocker/r-ver from docker hub, not using the one in the PR. 

To test, I'm using a dummy Dockerfile that basically uses your proposal in #754 and adds the extra script tests:

```
FROM ubuntu:jammy

LABEL org.opencontainers.image.licenses="GPL-2.0-or-later" \
      org.opencontainers.image.source="https://github.com/rocker-org/rocker-versioned2" \
      org.opencontainers.image.vendor="Rocker Project" \

ENV R_VERSION=4.3.2
ENV R_HOME=/usr/local/lib/R
ENV TZ=Etc/UTC

COPY scripts/install_R_source.sh /rocker_scripts/install_R_source.sh

RUN /rocker_scripts/install_R_source.sh

ENV CRAN=https://p3m.dev/cran/__linux__/jammy/latest
ENV LANG=en_US.UTF-8

COPY scripts /rocker_scripts

RUN /rocker_scripts/setup_R.sh

CMD ["R"]

# TESTING additional steps after r-ver steps:
RUN /rocker_scripts/install_jupyter.sh
RUN /rocker_scripts/install_verse.sh

```



Anyway, let's see if the above tests pass once the hotfixes you so nicely did reach DockerHub.

eitsupi:
@cboettig No, the Dockerfile used in the tests uses the latest scripts.

https://github.com/rocker-org/rocker-versioned2/blob/1828862223cd0a5357f2523c0dc85777c64fb9b7/tests/rocker_scripts/Dockerfile#L5-L10

The test case for not removing build deps is not covered here, so if CI goes green here, the `rocker/ml-verse` build will continue to fail.

eitsupi:
You should add the test case `install_texlive.sh` on `rocker/cuda:latest` to this file.

https://github.com/rocker-org/rocker-versioned2/blob/1828862223cd0a5357f2523c0dc85777c64fb9b7/tests/rocker_scripts/matrix.json

cboettig:
@eitsupi thanks for all your help here, sorry I don't always understand things.  


Yes, I see the test case copies the scripts, but it looks to me that is based on pulling r-ver from Dockerhub, and is not running install_R_source.sh script again.  Thus it appears to me the test case is grabbing the version of r-ver on Dockerhub that does not have builddeps purged.  We see here in the failing test, it pull from hub, copy the new scripts, and run only the `install_verse.sh`, right?: https://github.com/rocker-org/rocker-versioned2/actions/runs/7634519606/job/20799340714?pr=750#step:5:77  


I agree we need additional tests as well.  I think we actually have to test both `install_texlive.sh` and `install_jupyter.sh` in the cuda scenario (I think it is already covered in the r-ver case).  I'll try to add that test now.

eitsupi:
> and is not running install_R_source.sh script again. Thus it appears to me the test case is grabbing the version of r-ver on Dockerhub that does not have builddeps purged. We see here in the failing test, it pull from hub, copy the new scripts, and run only the `install_verse.sh`, right?

Correct. As you know, install_R_source.sh is quite time-consuming to execute, so regression tests are rarely triggered. The test is here, but it is not triggered in this PR.

https://github.com/rocker-org/rocker-versioned2/blob/1828862223cd0a5357f2523c0dc85777c64fb9b7/.github/workflows/r-build-test.yml#L56-L91

cboettig:
looks like the above test failures are network issues (HTTP 500 errors) on the manual install for texlive.  

Given how temperamental the network is on https://mirror.ctan.org/  and how large the cuda images are to begin with, I'm not sad to be getting texlive from the much more reliable ubuntu repos.  The changes in this PR should make the relevant scripts all more robust to working either with the hand-selected packages via tlmgr or the more standard (but larger) use of the official ubuntu apt binaries.  

For the reason we discussed above, these tests are still not 100% correct when running on the CI, because they are not running on the r-ver and cuda images that would be built here.  

I think a more complete check that I'm using locally is the above Dockerfile that runs these scripts in top of a r-ver rebuilt from scratch, and the same thing but for the cuda side.  I've tested both of those builds locally.  

We can tickle those checks again on the CI as well, but as a hotfix to correct the previous broken builds I think this will now succeed, modulo the stability of CTAN.org servers....

cboettig:
@eitsupi ok tests are green, including I think the additional test you mentioned. I think this is ready for review, thanks for all the help!

scripts/install_verse.sh
```diff
@@ -74,9 +74,13 @@ apt-get remove -y systemd
 apt-get -y autoremove
 
 ## Add LaTeX, rticles and bookdown support
-wget "https://travis-bin.yihui.name/texlive-local.deb"
-dpkg -i texlive-local.deb
-rm texlive-local.deb
+
```
eitsupi, line 77:
```suggestion
```

scripts/install_texlive.sh
```diff
@@ -1,4 +1,10 @@
...
+fi
```
eitsupi, line 6: I think this should after `set -e`

tests/rocker_scripts/matrix.json
```diff
@@ -17,7 +17,8 @@
     "install_python.sh",
     "install_jupyter.sh",
     "install_julia.sh",
-    "install_cuda-10.1.sh"
+    "install_cuda-10.1.sh",
+    "install_texlive.sh"
```
eitsupi, line 21: Is this necessary?
Isn't it tested via `install_verse.sh`?

tests/rocker_scripts/matrix.json
```diff
@@ -65,6 +66,12 @@
       "script_name": "install_tensorflow.sh",
       "script_arg": "none"
     },
+    {
+      "base_image": "rocker/cuda",
+      "tag": "cuda11.1",
```
eitsupi, line 71:
```suggestion
      "tag": "latest",
```

eitsupi (changes requested):

tests/rocker_scripts/matrix.json
cboettig: it is.  you said: 

> You should add the test case install_texlive.sh on rocker/cuda:latest to this file.

I'm not sure I entirely understand the syntax of matrix.json.  as you see below, I added this check using `install_texlive.sh`, and thought I had to list that here.

cboettig: @eitsupi ok, I've reverted this list back to what it was before.  I've kept the addition of adding a test on `rocker/cuda:latest`, for consistency with other tests I've just had it test against `install_verse.sh` which as you say will also test `install_texlive.sh` anyway.  Let me know what you think.

cboettig:
@eitsupi anything more on this? don't mean to rush you but do want to finish off this hot-fix for the broken builds  created by #743 , also current workflows won't run on some of the published images until this is merged.

cboettig:
I have had several workflows that are broken until this is fixed.  I am glad that we are respectful of the resources that Microsoft is providing for free, but I am also spending quite a bit of time working around the broken images at the moment and this probably impacts other users as well.

eddelbuettel:
> respectful of the resources that Microsoft is providing for free

It's complicated but I think "the community as a whole" is also overdoing it on builds, and CI runs, and whatnot.  (This comment is not about Rocker, or a stack, but more generic.  E.g. I have a beef with "kids these days" being lead to cookie-cutter development that runs each commit over a matrix of five or more builds for really no apparent reason.  My hill to to die on I guess.)

cboettig:
@eddelbuettel yeah I totally hear you that.  I think your point is that CI should be used thoughtfully, not mindlessly, otherwise developers and contributors time is not well spent.  I don't think anyone is worried that one of the 'magnificent 7' most profitable companies needs our help to ensure it remains financially viable.  There is of course a real carbon footprint to mindless CI (and needless use of chatbots etc), despite the aggressive carbon targets...

but I digress, I'm just hoping to get tensorflow working smoothly and other such bits of the python+gpu experience on our image during this brief moment when I have a few cycles to invest on this before other things take over all my time again.

eddelbuettel:
As an (index fund owning hence indirect) MSFT shareholder I am not too worried about their profitability.  Free resources tend to be a cost of user acquisition that gets amortised in the long run over eventual fees.  

It's the carbon footprint from endlessly running jobs just because we can...

But yep thanks for making it easy to deploy gpus. I appreciate it too per my recent bug report.

PR #751, opened 2024-01-24
fix: the flag of purge build deps

eitsupi:
A hot fix for #743


cboettig:
apologies for missing that and not having a test.  Would probably have been better to have it purge unless the env var was defined rather than checking the value.  Thanks for fixing this so quickly!

PR #765, opened 2024-02-29
fix #763

### LINKED ISSUE #763: install_julia.sh doesn't install the latest version of Julia
This potential bug doesn't affect pre-built images. It only affects users who want to call the script `install_julia.sh` to install the latest version of Julia in the Docker image.

If a user doesn't specify the environment variable `JULIA_VERSION`, the current script `install_julia.sh` will always install Julia v1.9.4, even though the latest stable version of Julia is 1.10.1 now. 

Changing this [line](https://github.com/rocker-org/rocker-versioned2/blob/7a2a65a498c2bdac90192c7c081fbfd7f86e775a/scripts/install_julia.sh#L40)
`latest_version <- sort(versions[is_stable], decreasing = TRUE)[1]`
to
`latest_version <- as.character(sort(numeric_version(versions[is_stable]), decreasing = TRUE)[1])`
should fix the issue.

Hua-Zhou:


eitsupi:
Thanks, looks like it is working well so we can merge this, but would you like to rewrite it to use juliaup?

eitsupi:
CI failures seem unrelated this change, so merging.

PR #768, opened 2024-03-08
update package list

cboettig:


scripts/install_geospatial.sh
```diff
@@ -72,8 +76,6 @@ install2.r --error --skipmissing --skipinstalled -n "$NCPUS" \
     terra \
     tidync \
     tmap \
-    geoR \
-    geosphere \
```
eitsupi, line 76: Removing the R packages will affect previous R version images, is that as intended?

PR #773, opened 2024-03-16
Ensure SIGTERM leads to graceful termination in rocker/binder

### LINKED ISSUE #771: SIGTERM isn't propegated or respected, making `docker stop` slow and then forceful
## Issue

`docker stop` and `kubectl delete` etc will pass a SIGTERM signal to kindly ask that it self-terminates, and after a while (10 seconds for docker, 30 seconds for kubectl), it will pass SIGKILL forcefully terminating everything. What often can happen is that a SIGTERM signal is only passed to process id 1 in the container, but then not propegated onwards to other processes that otherwise would terminate as long as they received the signal.

```shell
# is the container able to terminate on SIGTERM in less than 5 seconds
# or is it going to be killed forcefully after 10 seconds as is the default for
# docker to wait?
container_id=$(docker run -d rocker/binder)
timeout 5 docker stop $container_id && echo 'Stopped in time' || echo 'SIGTERM likely not respected'
```

## A resolution

By using an entrypoint called [`tini`](https://github.com/krallin/tini), we can resolve this.

```Dockerfile
FROM rocker/binder:latest

USER root
RUN apt-get update \
 && apt-get install -y --no-install-recommends \
        tini
USER ${NB_USER}

ENTRYPOINT ["tini", "--"]
```

```shell
# test of the Dockerfile above to conclude it can stop within a second
docker build . -t rocker/binder:tini

container_id=$(docker run -d rocker/binder:tini)
timeout 1 docker stop $container_id && echo 'Stopped in time' || echo 'SIGTERM likely not respected'
3362bb73d6610efc08fe89c72e46339f9045c9aa929b92dea49e8d62961c1381
Stopped in time
```

consideRatio:
Fixes #771, not by introducing `tini`, but by ensuring we don't let `sh` be the init process which in turn runs `jupyter lab` without propegating SIGTERM to it.

There may be a point to introduce `tini` or similar, but from the SIGTERM perspective, this does the trick as well.

This PR is modelled on for example https://github.com/rocker-org/rocker-versioned2/pull/740/files where I see that all Dockerfiles are updated along with a repsective .json file.

### DIFF
--- a/dockerfiles/binder_4.0.0.Dockerfile
+++ b/dockerfiles/binder_4.0.0.Dockerfile
@@ -13,7 +13,7 @@ RUN /rocker_scripts/install_jupyter.sh
 
 EXPOSE 8888
 
-CMD ["/bin/sh", "-c", "jupyter lab --ip 0.0.0.0 --no-browser"]
+CMD ["jupyter", "lab", "--ip", "0.0.0.0", "--no-browser"]
 
 USER ${NB_USER}
 
--- a/dockerfiles/binder_4.0.1.Dockerfile
+++ b/dockerfiles/binder_4.0.1.Dockerfile
@@ -13,7 +13,7 @@ RUN /rocker_scripts/install_jupyter.sh
 
 EXPOSE 8888
 
-CMD ["/bin/sh", "-c", "jupyter lab --ip 0.0.0.0 --no-browser"]
+CMD ["jupyter", "lab", "--ip", "0.0.0.0", "--no-browser"]
 
 USER ${NB_USER}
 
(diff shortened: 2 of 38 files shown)


cboettig:
@eitsupi thoughts on this?

consideRatio:
I did a force push. I saw that I had updated .json files incorrectly to reflect the changes in the Dockerfiles (left `/bin/sh -c` in the .json representation still)

eitsupi (approved):
Thank you for working on this.

I don't have anything to say, but please merge if you are sure because I have no idea if this Dockerfile will work correctly.

consideRatio:
I've tested to startup the Dockerfile built just adjusting CMD as this PR does and it works fine, but I don't understand what I changed in the .json files.

eitsupi:
I have confirmed that the JSON file has now been changed without any problems.

consideRatio:
> I don't have anything to say, but please merge if you are sure because I have no idea if this Dockerfile will work correctly.

@cboettig you hold the ball to get this merged - ok to go?

cboettig:
thanks for the ping!

Issue #779, opened 2024-04-06
Incorporation of Ubuntu LTS security patches - policy and timing


gregfrog:
I've been looking at the timing of Ubuntu LTS security patches being published, those patches getting into Ubuntu docker images and their incorporation here.  

What is the lead time in getting an Ubuntu LTS image with a security patch applied into the code here?  Is there a policy relating to that or is it a side-effect on other activities?  Some of the images based on actively maintained Ubuntu versions seem to be fairly old (r-ver 4.1 for example based on Focal seems not to have been rebuilt in a year but focal seems to have been patched since it was built), some are recent.  

Is there a policy describing which images are actively maintained and likely to pick up security patches and which are not?  I can't see one on the website or the wiki.

eitsupi:
See #614

eitsupi:
The next update will isolate and display tags that are no longer used on the wiki home. (#780)

Issue #776, opened 2024-03-24
Repository structure overhaul


eitsupi:
I will be working over the next few weeks on the repository overhaul that was discussed in several issues.

The current repository structure was originally designed to be updated manually, and I forced it to be changed to be updated automatically several years ago (#181), so it is not optimized for automatic updates. (Files that should be updated manually and files that are automatically generated are not completely separated.)
And, because Dockerfile is generated from json and R scripts, it is not possible to generate complex Dockerfiles that incorporate multi-stage builds.

Therefore, rewrite the structure assuming automation as follows:

- Dockerfile and bake files are generated by pouring variables into manually written templates with a template engine (The `glue` R pakcage).
- Variables are stored in TSV and JSON files and are automatically generated daily by CI.
- Completely rewrite the Dockerfiles to share layers instead of hierarchy.
  Now:
  ```dockerfile
  FROM rocker/r-ver:4.3.3
  RUN ...
  ```
  Future:
  ```dockerfile
  FROM ubuntu:jammy
  COPY rocker_scripts/install_R.sh /rocker_scripts/install_R.sh
  RUN /rocker_scripts/install_R.sh  # <- same as `rocker/r-ver`
  RUN ...
  COPY rocker_scripts/ /rocker_scripts/
  ```

PoC can be found here.
<https://github.com/eitsupi/rocker-versioned-next>

cboettig:
All for removing latest-daily.  why exclude rocker/binder though?

eitsupi:
> why exclude rocker/binder though?

As I wrote here, I think the current build strategy is not ideal for `rocker/binder`.
https://github.com/rocker-org/rocker-versioned2/issues/614#issuecomment-2028107573

cboettig:
:100: right, sorry I got confused between the two threads.  I've wondered if we should be pinning python versions in rocker/binder, but the rationale there given the recent security patch etc makes sense, and easier to do this with the build strategy you suggest in https://github.com/rocker-org/rocker-versioned2/issues/614#issuecomment-2028107573 !

eitsupi:
I forgot to close it.

PR #780, opened 2024-04-06
docs(wiki): split the image table to current and past

eitsupi:
Related to #776, #614, #779.

List images which the definition file has already been deleted in a separate table.

Looks like this:

image


cboettig:
yup, deprecating the 18.04 images looks good.

PR #782, opened 2024-04-21
BREAKING: repository structure change

eitsupi:
Part of #776

- [x] Update Dockerfiles
  - [x] Update main Dockerfiles
  - [x] Update devel Dockerfiles
  - [x] Update binder Dockerfiles
  - [x] Update cuda Dockerfiles
  - [x] Update extra Dockerfiles
- [x] Update docker-bake.json
  - [x] Update main docker-bake.json (renamed from core)
  - [x] Update devel docker-bake.json
  - [x] Update binder docker-bake.json (renamed to extra)
  - [x] Update cuda docker-bake.json (renamed to extra)
  - [x] Update extra docker-bake.json (renamed to experimental)
- [x] Update documents
- [x] Update CI
  - [x] Update Makefile
  - [x] Update GHA yaml files
  - [x] Update reports/wiki templates


eitsupi:
@cboettig Sorry for the delay, but now that we can generate at least some Dockerfiles and bakefiles, could you please take a look at this?

build/scripts/generate-args.R
```diff
@@ -0,0 +1,262 @@
...
+    dplyr::mutate(
+      r_major_latest = dplyr::if_else(dplyr::row_number() == dplyr::n(), TRUE, FALSE),
+      .by = r_major_version
+    ) |>
+    dplyr::select(!c(r_minor_version, r_major_version)) |>
+    # Supports the latest two patch versions and the latest two minor versions.
+    dplyr::filter(
+      r_minor_latest | dplyr::lead(r_major_latest, default = FALSE)
+    ) |>
+    dplyr::slice_tail(n = 3) |>
```
eitsupi, line 187: It should be changed so that it is not updated except for the two most recent versions.
Perhaps we need to change the file deletion to be determined by the R version rather than by the `rm` command as it is currently done for Dockerfiles.

build/scripts/generate-main-bakefiles.R
```diff
@@ -0,0 +1,204 @@
...
+    if (isTRUE(use_latest_tag)) {
+      .tags <- c(.tags, outer_paste(base_name, ":", latest_tag))
+    }
+  }
+
+  .tags
+}
+
+
+df_args <- fs::dir_ls(path = "build/args", regexp = r"((\d+\.){3}json)") |>
```
eitsupi, line 187: Should be filtered to latest two versions.

build/scripts/generate-main-dockerfiles.R
```diff
@@ -0,0 +1,71 @@
...
+            .close = "}}",
+            .trim = FALSE
+          )
+        )
+      }
+    )
+}
+
+
+df_args <- fs::dir_ls(path = "build/args", glob = "*.json") |>
```
eitsupi, line 44: Ditto.

build/scripts/generate-args.R
cboettig: agreed

cboettig (commented):
I think everything here so far looks very solid!  

Regarding the building last 2 versions: this all makes sense.  But in this new setup is it still possible to occasionally build older versions in order to benefit from the ubuntu-lts security patches released for them?

eitsupi:
> But in this new setup is it still possible to occasionally build older versions in order to benefit from the ubuntu-lts security patches released for them?

This is only possible for images that have the exact same structure as the most recent images.

For example, r-ver is fine for all versions, but rstudio cannot go back beyond a certain point because it becomes impossible to build linux/arm64 platform at some point.

build/scripts/generate-main-bakefiles.R
eitsupi: Done in 72c1e95

build/scripts/generate-main-dockerfiles.R
eitsupi: Done in 72c1e95

eitsupi:
@cboettig Do you think geospatial-ubuntugis is still needed?

cboettig:
>  Do you think geospatial-ubuntugis is still needed?

nope, let's drop this one!

eitsupi:
Since ml and ml-verse keep failing to build (and of course failed today), the latest version still seems to be stuck at R 4.3.2.

image
image

I think we need to finish and merge this PR soon and change it to do multi-stage builds and streamline it.......

cboettig:
> Since ml and ml-verse keep failing to build (and of course failed today), the latest version still seems to be stuck at R 4.3.2.

Looks like these are all due to inadequate space on the Github runner, right? (e.g. https://github.com/rocker-org/rocker-versioned2/actions/runs/8816018461/job/24199213471#step:7:4638 , `no space left on device`).  Like you say, hopefully we can avoid that with the new build strategy.

eitsupi:
> Looks like these are all due to inadequate space on the Github runner, right?

Yes, as you noted somewhere perhaps this is related to the number of parallels, but I don't know what the limitation is.
It is the cuda images that always fail, so perhaps when we parallelize, the spare storage is also split?

In any case, the number of parallelism is reduced in this PR to begin with, and even if the build fails, there will be no problem using the cache to repeat the build again (nothing will be changed except the failed image).

nathanweeks:
> > Since ml and ml-verse keep failing to build (and of course failed today), the latest version still seems to be stuck at R 4.3.2.
> 
> Looks like these are all due to inadequate space on the Github runner, right? (e.g. https://github.com/rocker-org/rocker-versioned2/actions/runs/8816018461/job/24199213471#step:7:4638 , `no space left on device`). Like you say, hopefully we can avoid that with the new build strategy.

In case a stop-gap solution might be helpful, the GitHub-hosted runners have a second block storage device with a mostly-unused file system mounted at `/mnt`:
```
runner@fv-az1432-846:~$ lsblk
NAME    MAJ:MIN RM  SIZE RO TYPE MOUNTPOINTS
loop0     7:0    0 63.9M  1 loop /snap/core20/2182
loop1     7:1    0   87M  1 loop /snap/lxd/27948
loop2     7:2    0 39.1M  1 loop /snap/snapd/21184
sda       8:0    0   75G  0 disk
└─sda1    8:1    0   75G  0 part /mnt
sdb       8:16   0   75G  0 disk
├─sdb1    8:17   0 74.9G  0 part /
├─sdb14   8:30   0    4M  0 part 
└─sdb15   8:31   0  106M  0 part /boot/efi
runner@fv-az1432-846:~$ df -h
Filesystem      Size  Used Avail Use% Mounted on
/dev/root        73G   53G   21G  72% /
tmpfs           7.9G  172K  7.9G   1% /dev/shm
tmpfs           3.2G  1.1M  3.2G   1% /run
tmpfs           5.0M     0  5.0M   0% /run/lock
/dev/sdb15      105M  6.1M   99M   6% /boot/efi
/dev/sda1        74G  4.1G   66G   6% /mnt
tmpfs           1.6G   12K  1.6G   1% /run/user/1001
```

A job step that relocates the the docker root directory to this "spare" file system could be added at the beginning of a job; e.g., containing commands like the following:

```
sudo sed -i 's#^{#{"data-root":"/mnt/docker",#' /etc/docker/daemon.json 
sudo systemctl restart docker
```

*Disclaimer:* I've only tested this approach interactively on a runner, where I verified `docker pull` worked & used /mnt/docker.

cboettig:
nice, all this looks good to me.  looking forward to this new setup.

cboettig:
@nathanweeks many thanks for this tip!  I can confirm this works on other builds I've tried with very large images, e.g.
https://github.com/boettiger-lab/k8s/actions/runs/8857066532/job/24324087762

eitsupi:
Let's merging for now...

shug0131:
I forked from this repo about a year ago https://github.com/shug0131/rocker-versioned2/ .  This basically adds a layer on top of rocker with extra packages, plus adds some extra libraries to allow the basic UAT to pass.  But it is anchored to the old version.  How would you recommend I go about merging it back with your changes to the repo structure? Would there be a way to avoid forking and making my repo copy-edit  directly from the main repo ?

eitsupi:
> How would you recommend I go about merging it back with your changes to the repo structure? Would there be a way to avoid forking and making my repo copy-edit directly from the main repo ?

I do not understand the intent of your question.
I don't think we can adopt the changes because the previous repository structure would not be able to generate the Dockerfiles as they are now.

PR #791, opened 2024-04-27
free space on runner?

cboettig:
just testing a few ideas for freeing build space in the current setup.


mskyttner:
I have been using this step [jlumbroso/free-disk-space@main](https://github.com/KTH-Library/kontarion/blob/master/.github/workflows/push-kontarion.yml#L17-L25) to free space on the runner successfully for a while when building an image which is based on ml-verse.

PR #792, opened 2024-04-29
ci(wiki): skip uploading wiki artifact in the default branch to speed up the action

eitsupi:
It seems very slow.


eitsupi:
The Ci failure is unrelated to this change. So merging.

PR #804, opened 2024-05-01
ci: fix caching of daily devel builds

eitsupi:


eitsupi:
The cache seems to be working well. There did not seem to be any updates to the R source today and nothing was built.
https://github.com/rocker-org/rocker-versioned2/actions/runs/8954208125
https://github.com/rocker-org/rocker-versioned2/actions/runs/8962175254

PR #805, opened 2024-05-01
fix: use gdebi-core to install RStudio

eitsupi:
python-setuptools is no longer available on Ubuntu 24.04.

### DIFF
--- a/scripts/install_rstudio.sh
+++ b/scripts/install_rstudio.sh
@@ -25,19 +25,8 @@ function apt_install() {
 
 apt_install \
     ca-certificates \
-    lsb-release \
-    file \
+    gdebi-core \
     git \
-    libapparmor1 \
-    libclang-dev \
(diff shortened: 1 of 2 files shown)


scripts/install_rstudio.sh
```diff
@@ -58,6 +51,11 @@ if [ "$UBUNTU_CODENAME" = "focal" ]; then
     UBUNTU_CODENAME="bionic"
 fi
 
+# TODO: remove this workaround for Ubuntu 24.04
+if [ "$UBUNTU_CODENAME" = "noble" ]; then
+    UBUNTU_CODENAME="jammy"
+fi
```
eitsupi, line 57: I believe this will need to be removed after a new version of RStudio is released.

eitsupi:
I think we can merge this now that it can pass tests with r-ver:latest (currently jammy) and r-ver:devel (currently noble).

PR #813, opened 2024-05-10
ci: need more space for this workflow

eitsupi:
https://github.com/rocker-org/rocker-versioned2/actions/runs/9031858606/job/24818930814#step:7:6417

```log
#60 exporting to image
#60 exporting layers
#60 exporting layers 93.4s done
#60 ERROR: mount callback failed on /tmp/containerd-mount3300003192: mount callback failed on /tmp/containerd-mount2529497484: failed to write compressed diff: failed to create diff tar stream: failed to copy: write /var/lib/buildkit/runc-overlayfs/content/ingest/693b8df03444e992ae900c05824dc3edea63723ab0d15dec82a99a5f9071aeb4/data: no space left on device
```

### DIFF
--- a/.github/workflows/extra.yml
+++ b/.github/workflows/extra.yml
@@ -56,6 +56,9 @@ jobs:
       matrix: ${{fromJson(needs.generate_matrix.outputs.matrix)}}
     steps:
       - uses: actions/checkout@v4.1.4
+      - name: Clean up
+        run: |
+          docker image prune --all --force
       - name: Login to Docker Hub
         uses: docker/login-action@v3
         with:


cboettig:
@eitsupi LGTM, feel free to merge.  

Do we want to add any of the additional solutions, e.g. in https://github.com/rocker-org/rocker-versioned2/pull/791 or mentioned in the threads?  I've been using the suggestion of using the extra volume as well and it works very nicely.  But if this is sufficient to get us going again it seems like a great starting point.

cboettig (approved):

eitsupi:
It works. https://github.com/rocker-org/rocker-versioned2/actions/runs/9039806681/job/24843185889

> Do we want to add any of the additional solutions, e.g. in #791 or mentioned in the threads? I've been using the suggestion of using the extra volume as well and it works very nicely. But if this is sufficient to get us going again it seems like a great starting point.

Yeah, I was also worried about whether this would be enough, but so far it seems to be fine.

PR #839, opened 2024-08-08
build: ensure only two minor versions (remove R 4.2.3 files)

### LINKED ISSUE #835: Are 4.2.3 Dockerfiles still autogenerated?
### Container image name

rocker/tidyverse:4.2.3

### Container image digest

none

### What operating system related to this question?

macOS

### System information

- Docker Desktop 4.32.0 (157355)
- MacOS 14.5
- Macbook Air M2

### Question

The Dockerfiles for 4.2.3 did not get the fixes that were applied to other Dockerfiles after #782, so they're now missing the following lines that are present for other releases:
```
ENV LANG=en_US.UTF-8
COPY scripts/bin/ /rocker_scripts/bin/
[...]
COPY scripts/install_s6init.sh /rocker_scripts/install_s6init.sh
COPY scripts/default_user.sh /rocker_scripts/default_user.sh
COPY scripts/init_set_env.sh /rocker_scripts/init_set_env.sh
COPY scripts/init_userconf.sh /rocker_scripts/init_userconf.sh
COPY scripts/pam-helper.sh /rocker_scripts/pam-helper.sh
```

I noticed this because the missing `COPY scripts/bin/ ...` leads to a build failure in all the tidyverse-containing images, because install_tidyverse.sh needs the Rocker version of `install2.r` for its `--skipmissing` option.

Looks like the difference emerged because `generate-dockerfiles.R` [only updates the latest 2 releases plus devel](https://github.com/rocker-org/rocker-versioned2/blob/f7161ec6d4310518df14a5ab47fdde098c8764fb/build/scripts/generate-dockerfiles.R#L51-L52), so 4.2.3[*] is skipped. 

Is the intended behavior here that all releases older than the last two will revert to being only manually edited? If yes, I can PR a hand-edit of the affected 4.2.3 dockerfiles and a patch to `dockerfiles/README.md` that explains only some files are autogenerated -- can you advise on the right phrasing? Would it also be good to annotate the relevant files as something like "no longer updated automatically"?

[*]Since the release of R 4.4.1, 4.3.3 is skipped as well, but none of the 4.3.3 dockerfiles (yet) change if I force regenerate them.

eitsupi:
Fix #835

Originally intended to build only the last two minor versions (which are known to be tested in CRAN and I believe are often considered supported R versions), we actually continued to build R4.2 after the R4.4 release.

The build script has been corrected and the R4.2 files will be correctly deleted and will no longer be built.


build/scripts/generate-matrix.R
```diff
@@ -22,7 +22,7 @@ supported_versions <- fs::dir_ls(path = "build/args", regexp = r"((\d+\.){3}json
   ) |>
   dplyr::slice_tail(n = 2, by = c(major, minor)) |>
   dplyr::filter(
-    minor == dplyr::last(minor) | patch >= dplyr::lead(patch)
+    minor == dplyr::last(minor) | dplyr::lead(minor) == dplyr::last(minor) & patch >= dplyr::lead(patch)
```
cboettig, line 25: nice! sorry I also overlooked this earlier!

Issue #857, opened 2024-09-17
Custom definition of `R_LIBS`


michaelmayer2:
### Container image name

rocker/r-ver:4.4.0 

### Container image digest

sha256:73adabf378930f8744f56bac1fe5e99870532eb4ced1a15075991bac909d21f3

### What operating system related to this question?

Linux

### System information

- Docker Version 27.1.1

### Question

I am curious why you added https://github.com/rocker-org/rocker-versioned2/blob/92e20d2bc2e6aed495a7cc359154627c7a9b048d/scripts/install_R_source.sh#L149 to `Renviron.site`. According to the R documentation, the presence of `site-library` in `${R_HOME}/site-library` alone will add this to `libPaths()` (cf. https://stat.ethz.ch/R-manual/R-devel/library/base/html/libPaths.html "First, .Library.site is initialized from R_LIBS_SITE. If this is unset or empty, the ‘site-library’ subdirectory of R_HOME is used. Only directories which exist at the time of initialization are retained. Then, .libPaths() is called..."). 

I have recently seen a case where the current definition of `R_LIBS` in `Renviron.site` in the docker image leads to the behaviour of `.libPaths()` listing the folders in reverse order (i.e. `R_LIBS_USER` the last) which makes package installations using `install.packages()` impossible (unless you really set target `.libPaths()` manually. 

I am sure there is valid reason for your approach to `Renviron.site` - just curious to understand the history of it.

michaelmayer2:
Steps to reproduce:

```
docker run -ti rocker/r-ver:4.4.0 bash
```
In container, run 
```
useradd -s /bin/bash -m mm 
su - mm
R
```

In R console, run 
```
install.packages("SuppDists")
.libPaths()
```

`.libPaths()` will report 
```
> .libPaths()
[1] "/usr/local/lib/R/site-library"             
[2] "/usr/local/lib/R/library"                  
[3] "/home/mm/R/x86_64-pc-linux-gnu-library/4.4"
```

After removing `Renviron.site` and rerunning a new R session, `libPaths()` will report 

```
> .libPaths()
[1] "/home/mm/R/x86_64-pc-linux-gnu-library/4.4"
[2] "/usr/local/lib/R/site-library"             
[3] "/usr/local/lib/R/library"  
```

as expected.

eitsupi:
@cboettig Any idea?

cboettig:
[...]

PR #866, opened 2024-10-22
remove custom definition of R_LIBS in scripts/install_R_source.sh

michaelmayer2:
As per discussion in https://github.com/rocker-org/rocker-versioned2/issues/857 this is a PR to fix the issue.

### DIFF
--- a/scripts/install_R_source.sh
+++ b/scripts/install_R_source.sh
@@ -145,9 +145,6 @@ mkdir -p "${R_HOME}/site-library"
 chown root:staff "${R_HOME}/site-library"
 chmod g+ws "${R_HOME}/site-library"
 
-## Fix library path
-echo "R_LIBS=\${R_LIBS-'${R_HOME}/site-library:${R_HOME}/library'}" >>"${R_HOME}/etc/Renviron.site"
-
 ## Clean up from R source install
 cd ..
 rm -rf /tmp/*


scripts/install_R_source.sh
```diff
@@ -145,9 +145,6 @@ mkdir -p "${R_HOME}/site-library"
 chown root:staff "${R_HOME}/site-library"
 chmod g+ws "${R_HOME}/site-library"
 
-## Fix library path
-echo "R_LIBS=\${R_LIBS-'${R_HOME}/site-library:${R_HOME}/library'}" >>"${R_HOME}/etc/Renviron.site"
```
eitsupi, line 149: Sorry I haven't caught up with the discussion, but do we need to remove this instead of changing the order?

eddelbuettel: I also fear (based on mere eyeballing) that the current definition may be quite right.  It's not easy to get this right for all cases.  I also think R prefers use of R_LIBS_SITE.  In Renviron (for the Debian package) I now set

```sh
# edd Apr 2003  Allow local install in /usr/local, also add a directory for
#               Debian packaged CRAN packages, and finally the default dir
# edd Jul 2007  Now use R_LIBS_SITE, not R_LIBS
# edd Jan 2023  Modify R_LIBS_{USER,SITE} respecting existing values (cf discussion with Kurt Hornik)
R_LIBS_USER=${R_LIBS_USER:-'%U'}
R_LIBS_SITE=${R_LIBS_SITE:-'/usr/local/lib/R/site-library:%S'}
```

and there also is this in Renviron.site

```sh
## # We comment out this line in the default Renviron, you can re-enable it here
## # @R_PLATFORM@ can be eg x86_64-pc-linux-gnu, see /etc/R/Renviron for R_PLATFORM
## # @MAJ_MIN_VERSION@ is likely something like '3.4', see /etc/R/Renviron too
## #
## # Also note that you set this per-user via '%p' and '%v' for platform and version
#R_LIBS_USER=${R_LIBS_USER-'~/R/@R_PLATFORM@-library/@MAJ_MIN_VERSION@'}
```

as well as this possible interacting with what I have above:

```
## edd Apr 2003  Allow local install in /usr/local, also add a directory for
##               Debian packaged CRAN packages, and finally the default dir
## edd Jul 2007  Now use R_LIBS_SITE, not R_LIBS
## edd Mar 2022  Now in Renviron.site reflecting R_LIBS_SITE
R_LIBS_SITE="/usr/local/lib/R/site-library/:${R_LIBS_SITE}:/usr/lib/R/library"
```

michaelmayer2: I am not sure where does that leave us with regards to this PR: Do we agree or disagree to remove the `R_LIBS` definition as proposed by this PR ?

drkrynstrng: Based on my testing using R 4.4.2, changing this line from using R_LIBS to R_LIBS_SITE will also resolve the issue of the user's library path being set last:

```
echo "R_LIBS_SITE=\${R_LIBS_SITE:-'${R_HOME}/site-library:${R_HOME}/library'}" >> "${R_HOME}/etc/Renviron.site"
```

After that change, the user's library path will always be set first.

PR #874, opened 2024-11-13
fix: remove non-root user from the base library/ubuntu image

### LINKED ISSUE #871: User ID Change in R4.4.2/Ubuntu 24 Image: rstudio User Now Assigned UID 1001, Causing Incompatibility with R4.4.1
### Container image name

rocker/verse:4.4.2

### Container image digest

rocker/verse@sha256:730f1963feee28420bb7417658349d627a7a8fa55e906824093245ac82d56b36

Linux

### System information

- Docker version 27.3.1, build ce12230

### Bug description

Starting from R:4.4.2, the base Ubuntu images were upgraded from Ubuntu 22.04 to Ubuntu 24.04. I noticed that in this updated image, the user ID 1000 is now assigned to the `ubuntu` user, while the `rstudio` user has been assigned ID 1001.

This change has caused some containers in our setup to break. Previously, the default user on our host systems also used ID 1000, which allowed seamless bind mounts. Files written by the `rstudio` user inside the container would align with the host’s default user ID, simplifying permissions management.

To adapt to this new setup, we would need to enable Docker's User Namespace Remapping feature, which requires making several adjustments on each of our hosts, along with elevated privileges (sudo access).

I’m uncertain if this shift in user IDs was intentional. Should we consider this a permanent change, or would it be adjusted in future releases? Knowing this would help us determine if we should update our configuration scripts accordingly. 
Thank you!

```sh
❯ docker run rocker/verse:4.4.1 id -u rstudio
1000
❯ docker run rocker/verse:4.4.2 id -u rstudio
1001
```

eitsupi:
Fix #871

### DIFF
--- a/build/templates/dockerfiles/cuda.Dockerfile.txt
+++ b/build/templates/dockerfiles/cuda.Dockerfile.txt
@@ -1,3 +1,5 @@
+# syntax=docker/dockerfile:1
+
 FROM nvidia/cuda:11.8.0-cudnn8-devel-ubuntu22.04
 
 ENV R_VERSION="{{r_version}}"
@@ -12,7 +14,12 @@ ENV LANG=en_US.UTF-8
 
 COPY scripts/bin/ /rocker_scripts/bin/
 COPY scripts/setup_R.sh /rocker_scripts/setup_R.sh
-RUN /rocker_scripts/setup_R.sh
+RUN <<EOF
+if grep -q "1000" /etc/passwd; then
+    userdel --remove "$(id -un 1000)";
+fi
+/rocker_scripts/setup_R.sh
+EOF
 
 CMD ["R"]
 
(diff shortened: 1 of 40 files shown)


eitsupi:
I will merge in time for the daily build.

pietertjerkstra:
How to use alternative users after this fix? We use the container with multiple users and a mounted volume to the /home folder. Since every user has its own username, she/he gets its own folder for saving user preferences and credentials. The preferences are saved on the mounted volume and are loaded the next time of usage. How to recreate this situation after this fix?

eitsupi:
I don't understand your question. I don't think it has anything to do with this change.
We recommend asking questions on Stack Overflow, etc.

pietertjerkstra:
Thank you for your response, the question is posted here: [stackoverflow](https://stackoverflow.com/questions/79357865/rocker-rstudio-username)

eitsupi:
I really don't know what you are trying to achieve, but if you want to change the username, the steps are described in the documentation in this repository.
https://github.com/rocker-org/rocker-versioned2/blob/master/NEWS.md#2024-12

pietertjerkstra:
Thank you very much, that is indeed what I try to do. I will try the method proposed in your link

PR #891, opened 2024-12-18
refactor!: remove the feature to change the default non-root user name via `USER` env var

### LINKED ISSUE #489: chown when specifying username in rstudio container
Customizing the username produces a long-running (more than 15 minutes, have not waited until the end) chown process before the rstudio server is started. This appears to be related to the volumes added to the container. Keeping the default user "rstudio" does not cause a problem. The workaround is obviously to just keep the default user which is fine for local insecure containers. Enabling authentication, adding a password, and changing the username would be ideal for cloud deployment but this will not be feasible if using large volumes.

Docker on WSL2 on Windows 10
Host username is huser
Container username is duser

Setup up cifs volume then run container with both local volume and network volume

```
docker volume create \
  --driver local \
  --opt type=cifs \
  --opt device=//xxx.xxx.xxx.xxx/Array \
  --opt o=username=huser,password=pswd,uid=1000 \
  --name Array

docker run -d \
    --name Rocker \
    --hostname Rocker \
    --network bridge \
    -e USER=duser \
    -e DISABLE_AUTH=true \
    -p 127.0.0.1:8787:8787 \
    -v /mnt/c/Users/huser/local:/home/duser/local \
    -v Array:/home/duser/array \
    rocker/rstudio:latest
```

I am relatively new to docker so I apologize if this is a docker problem rather than an image problem.

### LINKED ISSUE #889: Addressing the Time-Consuming chown -R Command in scripts/init_userconf.sh
### Container image name

rocker/rstudio:4.0.0

### Container image digest

_No response_

Linux

### System information

- Rocky Linux 8.10

### Bug description

When starting a container, instead of using the default user, I configure a different user and have encountered an issue: if the home directory contains many subdirectories, starting the container becomes very time-consuming. The reason for this is that the chown -R command in scripts/init_userconf.sh (https://github.com/rocker-org/rocker-versioned2/blob/master/scripts/init_userconf.sh#L174) is very time-intensive. In another issue(https://github.com/rocker-org/rocker-versioned2/issues/489), others suggested abandoning the functionality to modify the user, but I don't think we need to discard this feature. Is it possible to simply remove the chown -R command and let the user ensure the permissions of the home directory themselves? Alternatively, could we add a new environment variable that triggers the chown -R command when set?

_No response_

eitsupi:
Fix #489
Close #889

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,11 @@
 # News
 
+## 2024-12
+
+### Changes in pre-built images
+
+The feature to change the name of the default non-root user via the `USER` environment variable has been removed.
+
 ## 2024-04
 
 The repository has been completely reorganized.
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,8 +4,21 @@
 
 ### Changes in pre-built images
 
-The feature to change the name of the default non-root user via the `USER` environment variable has been removed.
-([#891](https://github.com/rocker-org/rocker-versioned2/pull/891))
+- The feature to change the name of the default non-root user via the `USER` environment variable has been removed.
+  ([#891](https://github.com/rocker-org/rocker-versioned2/pull/891))
+  If you want to change the default user name, you should creat a new user.
+  For example, something like this:
+
+  ```dockerfile
+  FROM rocker/rstudio:4
+  ENV DEFAULT_USER=new_user
+  RUN <<EOF
+  if grep -q "1000" /etc/passwd; then
+      userdel --remove "$(id -un 1000)";
+  fi
+  /rocker_scripts/default_user.sh
+  EOF
+  ```
 
 ## 2024-04
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,7 @@
 ### Changes in pre-built images
 
 The feature to change the name of the default non-root user via the `USER` environment variable has been removed.
+([#891](https://github.com/rocker-org/rocker-versioned2/pull/891))
 
 ## 2024-04
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,7 +5,7 @@
 ### Changes in pre-built images
 
 - The feature to change the name of the default non-root user via the `USER` environment variable has been removed.
-  ([#891](https://github.com/rocker-org/rocker-versioned2/pull/891))
+  ([#891](https://github.com/rocker-org/rocker-versioned2/pull/891))\
   If you want to change the default user name, you should creat a new user.
   For example, something like this:
 
--- a/scripts/init_userconf.sh
+++ b/scripts/init_userconf.sh
@@ -3,7 +3,7 @@
 
 ## Set defaults for environmental variables in case they are undefined
 DEFAULT_USER=${DEFAULT_USER:-rstudio}
-USER=${USER:=${DEFAULT_USER}}
+USER=${DEFAULT_USER}
 USERID=${USERID:=1000}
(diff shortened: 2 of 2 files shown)


cboettig (approved):
Nicely streamlined here, this makes sense to me!

rmjstn:
Hello,

The modification `USER=${DEFAULT_USER}` vs `USER=${USER:=${DEFAULT_USER}}` caused a problem. If I defined the USER in the Docker Compose file, it's ignored. USER is equal to DEFAULT_USER, and USERID is equals to that defined in the Docker Compose file. 

```
    echo "deleting the default user"
    userdel "$DEFAULT_USER"
    echo "creating new $USER with UID $USERID"
    useradd -m "$USER" -u "$USERID"
    mkdir -p "${USERHOME}"
    chown -R "$USER" "${USERHOME}"
    usermod -a -G staff "$USER"
```

Docker Compose file : 

```
    environment:
      - USER=remy
      - USERID=10042
```

Logs : 

```
deleting the default user
creating new rstudio with UID 10042
```

If I set  `USER=${USER:=${DEFAULT_USER}}`  in init_userconf.sh, it works for me.

Is there something I'm missing ?

Thanks,

eitsupi:
I'm sorry, but I don't understand what you are trying to say.
The purpose of this change is to disallow username changes using the `USER` environment variable.
In other words, the behavior you observed is as we intended and there is nothing wrong with it.

rmjstn:
Ha OK. Sorry, I misunderstood the purpose of the modification.
[This page](https://rocker-project.org/images/versioned/rstudio.html#environment-variables) had to explain how to value the USER variable.
I use this feature to map the user with the host one. I map the volume "/etc/shadow:/etc/shadow" for using the same password. So I've got this error while providing the USERID : 
```
creating new rstudio with UID 10042
useradd: failure while writing changes to /etc/shadow
```
I don't want to configure the password in the Docker Compose file.

Thanks

eitsupi:
> Ha OK. Sorry, I misunderstood the purpose of the modification. [This page](https://rocker-project.org/images/versioned/rstudio.html#environment-variables) had to explain how to value the USER variable. I use this feature to map the user with the host one. I map the volume "/etc/shadow:/etc/shadow" for using the same password. So I've got this error while providing the USERID :
> 
> ```
> creating new rstudio with UID 10042
> useradd: failure while writing changes to /etc/shadow
> ```
> 
> I don't want to configure the password in the Docker Compose file.

Thanks for the response. Okay, I understand your use case.

However, the use of this environment variable has been deprecated for the past several years.
See #490

If you prefer to use an arbitrary user name, build the image yourself as described here. (Perhaps this is worth adding to the documentation, which I will work on in due time if I remember)
https://github.com/rocker-org/rocker-versioned2/blob/master/NEWS.md#2024-12

rmjstn:
What I find disturbing (if I understand correctly the problem I encountered) is that the content of the rocker/verse:4.4.2 image has changed over time without being reversioned (4.4.2.1, 4.4.2.2...). If a server had an old cached image it worked. This means that the process is not reproducible.

eitsupi:
> What I find disturbing (if I understand correctly the problem I encountered) is that the content of the rocker/verse:4.4.2 image has changed over time without being reversioned (4.4.2.1, 4.4.2.2...). If a server had an old cached image it worked. This means that the process is not reproducible.

See <https://rocker-project.org/use/reproducibility.html>.

rmjstn:
Thanks for this clarification.

PR #905, opened 2025-02-08
refactor!: remove files for `rocker/cuda`, `rocker/ml`, and `rocker/ml-verse`

### LINKED ISSUE #903: Migrate cuda based images out of rocker-versioned2
I am fed up with the amount of questions about cuda and Python setup and the maintenance hassle, and strongly believe that users should install any version of Python using uv on the version of cuda image they want to use. (and then use rig to install and use any version of R)

The situation has changed dramatically from a few years ago when there was no rig or uv, and I think the significance of the old kind of pre-built image is declining.

@cboettig @noamross Thoughts?

eitsupi:
Close #903
Close #756


eitsupi:
https://github.com/rocker-org/rocker-versioned2/issues/903#issuecomment-2646881336

cboettig:
nice work tidying this all up!

PR #917, opened 2025-03-25
fix(install_quarto.sh): release option works

### LINKED ISSUE #916: Typo in install_quarto.sh for "release" option
The script checks `$RSTUDIO_VERSION" = "release"` instead of `$QUARTO_VERSION" = "release"`, so this option doesn't work:
https://github.com/rocker-org/rocker-versioned2/blob/abe4fc1f3fb1c50281f9de5b3636f8a117126f98/scripts/install_quarto.sh#L57

Because of the wrong if-else branch it doesn't grep the numeric version from _download.json, so QUARTO_VERSION remains "release" and it tries to download `https://github.com/quarto-dev/quarto-cli/releases/download/vrelease/quarto-release-linux-${ARCH}.deb` with a 404 error.

banfai:
Fixes #916


eitsupi (approved):
Thanks for your contribution!

PR #934, opened 2025-07-05
`quiet` and `verbose` options to `install2r`

zivankaraman:
I often build rocker-based images, which require installing many R packages, many of which require compilation. With the default installation options, this can produce very, very long log files that are impractical. The ‘`quiet`’ and ‘`verbose`’ options to ‘`install2r`’ that I added allow for fine-tuning the output quantity.
Please let me know if you require any additional details or clarification regarding these changes.


eddelbuettel:
Also, do you have a (small, pererably) package where your PR shows a difference in installation 'noise'?

(I have also been thinking to add default compiler flags to make eg RcppEigen-using compilation more quiet...)

eddelbuettel:
Also, do you have a (small, pererably) package where your PR shows a difference in installation 'noise'?

(I have also been thinking to add default compiler flags to make eg RcppEigen-using compilation more quiet...)

zivankaraman:
Thank you for your comments. My use case stemmed from building a rocker-based Docker image, so I modified the install2r included in the rocker repo. I will follow your suggestions.

eddelbuettel:
Yes, and coincidentally I just made essentially the same comment [in a reply to the littler issue just opened](https://github.com/eddelbuettel/littler/issues/124#issuecomment-3042256018) by @zivankaraman (and before I saw this).  We may want this here locally "in the meantime" and we want it "eventually as well" upstream so that it bubbles through the CRAN package process too.

Issue #950, opened 2025-09-23
Possible rocker/pharma_cr image


bms63:
### Container image name

rocker/pharma_cr image

### Container image digest

NA

### Question

I was just wondering about building a rocker image specific to pharma?  We have a lot of packages bespoke to our needs on the Clinical Reporting side.

If possibility - I am happy to start chasing.

bms63:
Sorry I am new to this world of docker/rocker.  I did not 

We have a lot of R packages created for pharma.  I have been using rocker in GHA and found them super helpful, but have to install the pharma packages.  

I was just hoping to promote this image as a lot of folks in pharma are using R.  Thought this could be useful like the tidyverse image.

eitsupi:
Thank you for clarifying that. It's very helpful to know the use case.

Installing packages quickly in R on Linux has become less difficult than before, and I believe it's quite simple if you prepare a DESCRIPTION file.
In fact, in this repository, we control the R packages used in CI/CD with the following DESCRIPTION file.
https://github.com/rocker-org/rocker-versioned2/blob/52f3e8e5ecfafb77a64c1479acbb97f743373f6c/build/DESCRIPTION

My recent thinking is that users should control which R packages they install and use themselves, and that it might not be ideal to provide pre-installed images with a large number of R packages from the start.

On the other hand, I recognize the popularity of rocker/tidyverse, and it's giving me a headache.

eddelbuettel:
You do have access to arm64 -- that is how I am building arm64 for r2u.  It's all in the open (i.e. you can see my yaml files too etc).  You need to modify you Dockerfile run a little to invoke qemu and do it twice, and afterwards all is pretty automatic and amazing.  See [their blog post](https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/) and more documentation (as always, it is there, along with enough other documentation to drown you).  

As for repo review, I am not sure I can do that.  You know your subject matter, and you will see when it works or croaks so I am sure you can take it from there.  Let a thousand containers bloom!

PR #951, opened 2025-09-25
init pharmaverse rocker

Polkas:
- [ ] decide which r packages to add
- [ ] decide if we want to include rstudio
- [ ] check if any pharmaverse package does not need any additional SYSTEM deps

### DIFF
--- a/bakefiles/4.5.0.docker-bake.json
+++ b/bakefiles/4.5.0.docker-bake.json
@@ -7,6 +7,7 @@
             "r-ver",
             "rstudio",
             "tidyverse",
+            "pharmaverse",
             "shiny",
             "shiny-verse",
             "verse",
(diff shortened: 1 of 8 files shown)


eitsupi:
Thank you for your interest in this repository, but wait, I don't want to add new images, so I can't welcome contributions like this.

If you've been looking at this repository lately, you'll see that we've actually been building fewer images in this repository, and I don't see any chance of increasing them.

eddelbuettel:
To add to this, please see #950 where this came up first (and opening an issue is the best way to gauge interest and applicability) and where we more or less recommended that this may be best suited to a downstream repo someone else (you maybe?) owns and runs.  This allows us to focus on what we do.

eddelbuettel:
Yes -- it is widely used, and there are a lot of packages. So it would make sense for someone with domain expertise to look after it!

And building 'derivative' images is easy!  I did just that when I worked with some singlecess use cases. This [blog post from 2023](https://dirk.eddelbuettel.com/blog/2023/08/13/) is a demo of working on top of what we do in Rocker (in my case more the r2u stack than tidyverse) to provide 150 or so packages.  As I say over in the r2u repo: Fast. Easy. Reliable.

It absolutely is a valid use case but @eitsupi and I have to me mindful of our time commitments.

PR #989, opened 2026-03-06
feat: enable linux/arm64 builds for rocker/geospatial

cboettig:
## Summary

- **Drop archived CRAN packages** from `install_geospatial.sh`: `RandomFields` (archived 2022), `rgdal`, `rgeos`, `maptools` (archived 2023, superseded by `sf`/`terra`). All remaining packages have arm64 binaries on P3M for Ubuntu Noble.
- **Rebase geospatial on tidyverse** instead of verse: remove the TeX Live layer (`install_verse.sh`, `install_texlive.sh`) from the Dockerfile template. LaTeX is not needed for geospatial workflows and was the architectural blocker for arm64 support.
- **Add `linux/arm64`** to geospatial platforms in the bakefile template and all versioned bakefiles (4.4.3, 4.5.2, 4.5.3).

## Test

Built and tested on a native arm64 machine. All key packages load correctly: `sf`, `terra`, `stars`, `spdep`, `hdf5r`, `lidR`, `tmap`, `sfarrow`, `ncdf4`, `geoR`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,22 @@
 # News
 
+## 2026-03
+
+### Breaking changes
+
+- `rocker/geospatial` no longer inherits from `rocker/verse`; it now builds on `rocker/tidyverse`.
+  TeX Live and publishing packages (`bookdown`, `rticles`, etc.) are no longer included.
+  Use `rocker/verse` or add `install_verse.sh` in a downstream image if LaTeX is needed.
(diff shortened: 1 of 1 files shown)


cboettig:
@eitsupi let me know what you think about this?  I've been getting a bunch of requests about arm support for geospatial (mostly for mac users I think).  

It now extends from tidyverse instead of verse, which i think is consistent with what you've also suggested elsewhere.

PR #1024, opened 2026-07-22
Map resolute to jammy for RStudio install

cboettig:
`ubuntu:latest` is now 26.04 LTS (resolute). RStudio publishes no resolute (or noble) deb — only jammy — so the devel build 404s daily on `install_rstudio.sh`. Remap resolute→jammy, matching the existing noble→jammy handling.

Verified: jammy deb installs and runs on a resolute container.

Fixes the daily R-devel rstudio build failure.


eitsupi:
Thanks, but it seems that ubuntu 26.04 not supported yet 🤔
https://dailies.rstudio.com/

